PR #7, opened 2022-05-13
Move SCIP to LSIF conversion into library.

varungandhi-src:
Mostly a no-op, exposing conversion functionality as a public API.

### Test plan

Existing tests are passing, double-checked that `src-cli` compiles fine
in https://github.com/sourcegraph/src-cli/pull/742 after using a replace
directive to point to my local `scip` repo with this patch.


varungandhi-src:
This is mostly a no-op, so I'm merging this without review.

PR #14, opened 2022-05-20
Add `scip snapshot` subcommand

olafurpg:
### Test plan

Manually ran `scip snapshot` against an example index. We're already using the same snapshot logic for reprolang tests so if we want to test this PR it would have to be by invoking the cli directly.

### DIFF
--- a/bindings/go/scip/symbol_formatter.go
+++ b/bindings/go/scip/symbol_formatter.go
@@ -13,6 +13,15 @@ type SymbolFormatter struct {
 	IncludeDescriptor     func(descriptor string) bool
 }
 
+// VerboseSymbolFormatter formats all parts of the symbol.
+var VerboseSymbolFormatter = SymbolFormatter{
+	IncludeScheme:         func(scheme string) bool { return true },
+	IncludePackageManager: func(_unused string) bool { return true },
+	IncludePackageName:    func(_unused string) bool { return true },
+	IncludePackageVersion: func(_unused string) bool { return true },
+	IncludeDescriptor:     func(_unused string) bool { return true },
+}
+
 // DescriptorOnlyFormatter formats only the descriptor part of the symbol.
 var DescriptorOnlyFormatter = SymbolFormatter{
 	IncludeScheme:         func(scheme string) bool { return scheme == "local" },
--- a/cmd/stats.go
+++ b/cmd/stats.go
@@ -1,0 +1,76 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"path/filepath"
+
+	"github.com/hhatto/gocloc"
+	"github.com/sourcegraph/scip/bindings/go/scip"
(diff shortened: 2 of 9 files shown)


bindings/go/scip/symbol_formatter.go
```diff
@@ -13,6 +13,15 @@ type SymbolFormatter struct {
 	IncludeDescriptor     func(descriptor string) bool
 }
 
+// VerboseSymbolFormatter formats all parts of the symbol.
+var VerboseSymbolFormatter = SymbolFormatter{
+	IncludeScheme:         func(scheme string) bool { return true },
+	IncludePackageManager: func(_unused string) bool { return true },
```
tjdevries, line 19: can't these just be `_` -- I'm not sure why you're calling them `unused` when they all are not used in this formatter.

cmd/stats.go
```diff
@@ -0,0 +1,76 @@
...
+	stats := &indexStatistics{
+		Documents:   int32(len(index.Documents)),
+		LinesOfCode: loc.Total.Code,
+		Occurrences: 0,
+		Definitions: 0,
+	}
+	for _, document := range index.Documents {
+		for _, occurrence := range document.Occurrences {
+			stats.Occurrences += 1
+			if occurrence.SymbolRoles&int32(scip.SymbolRole_Definition) > 0 {
```
tjdevries, line 57: side comment: do we have any `scip-util` (i hate util names, plz don't let me name it that) library already in Go so that we can just say `scip.isDefinition(occurent)`? Just wondering if we were already building that somewhere or not (or if it was part of this repo)

cmd/stats.go
```diff
@@ -0,0 +1,76 @@
...
+				stats.Definitions += 1
+			}
+		}
+	}
+	return stats, nil
+}
+
+func countLinesOfCode(index *scip.Index) (*gocloc.Result, error) {
+	root, err := url.Parse(index.Metadata.ProjectRoot)
```
tjdevries, line 66: should we check that this path exists locallly? I guess I'm assuming `gocloc` handles this as well though, so probably not worth it.

tjdevries (approved):
LGTM tho, you can ignore as many comments as you want

bindings/go/scip/symbol_formatter.go
olafurpg: Good point. I renamed  these into `_`

cmd/stats.go
olafurpg: I added a `SymbolRole.Matches(Occurrence) bool` helper method. Since we are in the same package, we can add methods on scip types.

cmd/stats.go
olafurpg: I added error handling for the case when the root directory doesn't exist. I assumed it was not a problem since we'll get an error that the underlying files don't exist either way.

olafurpg (commented):
Thank you for the review!

PR #15, opened 2022-05-21
Haskell bindings for SCIP

donsbot:
This uses google's proto-lens generator for Haskell. More details at:

https://github.com/google/proto-lens#readme

For full CI instructions see

https://github.com/donsbot/if-proto-scip-hs/blob/main/.github/workflows/ci.yml

PR #16, opened 2022-05-21
Add Haskell bindings for SCIP

olafurpg:
Supersedes #15 

### Test plan
See the CI go green.


olafurpg:
Ideally, we should automate the generation of the proto files with bufbuild but I'm tempted to leave this for a future step. This PR is already an improvement by adding Haskell instructions so I'm inclined to recommend we merge it as is and re-run the Haskell generation manually.

tjdevries:
Agreed. I think we can add automated build later (better to get the bindings in first, IMO)

.github/workflows/haskell.yml
```diff
@@ -0,0 +1,53 @@
...
+      image: ghcr.io/facebookincubator/hsthrift/ci-base:ghcup
+      options: --cpus 2 --security-opt=seccomp=unconfined
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install additional tools
+        run: |
+          apt-get update
+          apt install -y unzip
+        working-directory: bindings/haskell
```
varungandhi-src, line 24: nit: Bunch of these `working-directory: bindings/haskell` lines could be removed without changing the semantics.

bindings/haskell/README.md
```diff
@@ -0,0 +1,52 @@
+[![CI](https://github.com/donsbot/if-proto-scip-hs/actions/workflows/ci.yml/badge.svg)](https://github.com/donsbot/if-proto-scip-hs/actions/workflows/ci.yml)
+
+# Haskell bindings for SCIP
+
+These bindings use Google's proto-lens generator for Haskell. More details at:
+
+https://github.com/google/proto-lens#readme
```
varungandhi-src, line 7:
```suggestion
These bindings use [Google's proto-lens generator for Haskell](https://github.com/google/proto-lens).
```

bindings/haskell/if-proto-scip-hs.cabal
```diff
@@ -0,0 +1,21 @@
+cabal-version:      3.6
+
+name:               if-proto-scip-hs
```
varungandhi-src, line 3: Rename to `scip-haskell`?

bindings/haskell/if-proto-scip-hs.cabal
```diff
@@ -0,0 +1,21 @@
...
+maintainer:         code-intel@sourcegraph.com
+homepage:           https://github.com/sourcegraph/scip
+category:           Language
+build-type:         Simple
+
+library
+    ghc-options: -Wall
+    exposed-modules:
+        Proto.Scip,
+        Proto.Scip_Fields
```
varungandhi-src, line 18: Can we tweak the generation so that the names for the modules are `SCIP` and `SCIPFields`? That would match existing Haskell conventions better around acronyms.

bindings/haskell/if-proto-scip-hs.cabal
```diff
@@ -0,0 +1,21 @@
...
+    hs-source-dirs:   src
+    default-language: Haskell2010
+    build-depends: proto-lens-runtime, base
```
varungandhi-src, line 21: We should specify some version ranges here. ([examples](https://cabal.readthedocs.io/en/3.4/cabal-package.html))

We can probably add a freeze file (analogous to lock files elsewhere) later.

bindings/haskell/README.md
```diff
@@ -0,0 +1,52 @@
+[![CI](https://github.com/donsbot/if-proto-scip-hs/actions/workflows/ci.yml/badge.svg)](https://github.com/donsbot/if-proto-scip-hs/actions/workflows/ci.yml)
```
varungandhi-src, line 1: This link + 1 more link below need to be updated to point to this repo.

bindings/haskell/README.md
```diff
@@ -0,0 +1,52 @@
...
+curl --proto '=https' --tlsv1.2 -sSf https://downloads.haskell.org/~ghcup/$GHCUPVERSION/x86_64-linux-ghcup-$GHCUPVERSION > /usr/bin/ghcup && \
+chmod +x /usr/bin/ghcup
+ghcup install ghc 8.10.7 --set
+ghcup install cabal 3.6.0.0 --set
+```
+
+```sh
+# macOS
+GHCUPVERSION=0.1.17.4
+curl --proto '=https' --tlsv1.2 -sSf https://downloads.haskell.org/~ghcup/$GHCUPVERSION/x86_64-apple-darwin-ghcup-$GHCUPVERSION > /usr/local/bin/ghcup && \
```
varungandhi-src, line 26: suggestion: Maybe we should just recommend the official command?

```
curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
```

That is cross-platform (direct copy-pasting works for non x86_64 too), and will update for newer ghcup versions which are unlikely to break this basic functionality.

We already mention the CI script earlier, so one can look there for an exact version if desired.

bindings/haskell/README.md
```diff
@@ -0,0 +1,52 @@
+[![CI](https://github.com/donsbot/if-proto-scip-hs/actions/workflows/ci.yml/badge.svg)](https://github.com/donsbot/if-proto-scip-hs/actions/workflows/ci.yml)
+
+# Haskell bindings for SCIP
+
+These bindings use Google's proto-lens generator for Haskell. More details at:
+
+https://github.com/google/proto-lens#readme
+
+Building: see the following workflow file for most up to date:
```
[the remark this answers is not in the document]
varungandhi-src: Oh! I just realized. I'm not sure if this is needed here.

Instead, this should probably be part of the `dev/proto-generate.sh` script. We should update that script in this PR or a follow-up PR, and also update Development.md to mention that we have auto-generated Haskell bindings.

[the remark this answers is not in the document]
varungandhi-src: NVM, I just saw your and TJ's comments. 🤦 Marking this as resolved since this is temporary.

varungandhi-src:
I've filed https://github.com/sourcegraph/scip/issues/18 to incorporate Haskell generation into bufbuild.

.github/workflows/haskell.yml
```diff
@@ -0,0 +1,53 @@
...
+  ci:
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc: [8.10.7]
+    runs-on: ubuntu-latest
+    container:
+      # ubuntu with ghcup already installed
+      image: ghcr.io/facebookincubator/hsthrift/ci-base:ghcup
+      options: --cpus 2 --security-opt=seccomp=unconfined
```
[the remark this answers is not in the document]
varungandhi-src: Hmm, the commit message has some details:

> hhvm makes a couple of syscalls that are blocked by default on github's docker seccomp policy. We need to allow those. Unblocking all for now (see https://docs.docker.com/engine/security/seccomp/ )

I think we can disable this here since we aren't going to be using HHVM.

bindings/haskell/if-proto-scip-hs.cabal
varungandhi-src: I've manually renamed these for now.

varungandhi-src (approved):

varungandhi-src:
The Haskell CI job seems to be awfully slow. Just installing GHC is taking about a minute, and building the protoc plugin takes nearly 3 minutes. Yikes!

.github/workflows/haskell.yml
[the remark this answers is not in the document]
donsbot: It's just left over from the hhvm CI, which uses a ton of low level stuff. Not needed (and I updated https://github.com/donsbot/if-proto-scip-hs to reflect that). Thanks for checking!

bindings/haskell/README.md
donsbot: Yep, make sure to use the official ghcup (its the best way to get the correct version of things)

varungandhi-src:
Makes sense. I think this is okay for now, since schema changes are much less frequent compared to other changes, but something we may want to revisit later if we run into this often.

PR #23, opened 2022-05-24
ci: Give workflow jobs more useful names.

varungandhi-src:
The names show up in the UI, and having a bunch of jobs all
named 'check' makes it impossible to distinguish which job
is being referred to.

PR jobs as seen after merging

### Test plan

n/a

### DIFF
--- a/.github/workflows/golang.yml
+++ b/.github/workflows/golang.yml
@@ -9,7 +9,7 @@ on:
       - '**/go.sum'
 
 jobs:
-  check:
+  go-test:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


varungandhi-src:
Much better!

new job names from this PR

olafurpg:
Where do you see that view? I always see the name of the workflow first so the new output is more noisy

CleanShot 2022-05-25 at 10 54 58@2x

varungandhi-src:
After merging.

View details button

varungandhi-src:
Yeah, I know it is a bit more verbose pre-merge, but I figured that was not so bad. IDK why GitHub doesn't use the same names in both places. 🤷

olafurpg:
Aha, I had never used the "View details" button. It's surprising GH has different naming conventions for these jobs. I suppose they will iron out this quirk eventually. Totally fine by me to repeat the job name 👍

PR #27, opened 2022-05-24
Fix a bug related to textDocumentation/implementation

olafurpg:
Previously, doing "Find implementations" only showed one result when there were multiple results available. Now, all available results should be shown after this fix.

Opening this as a draft since it's still not working 100% and I'm gonna continue with this PR tomorrow.

### Test plan
See newly added snapshot tests. I'm also manually testing this change by uploading the generated dump.lsif to Sourcegraph here https://sourcegraph.com/github.com/sourcegraph/scip-typescript@34337f2fa29d3a83c71814458d19741ae4f21db9/-/blob/snapshots/input/syntax/src/inheritance2.ts?L2:3&subtree=true#tab=implementations_typescript

### DIFF
--- a/bindings/go/scip/convert.go
+++ b/bindings/go/scip/convert.go
@@ -92,7 +92,7 @@ func ConvertSCIPToLSIF(index *Index) ([]reader.Element, error) {
 type graph struct {
 	ID                   int
 	Elements             []reader.Element
-	symbolToResultSet    map[string]symbolInformationIDs
+	symbolToResultSet    map[string]*symbolInformationIDs
 	inverseRelationships map[string][]*Relationship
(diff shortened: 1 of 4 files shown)


olafurpg:
It's working correctly now 

CleanShot 2022-05-25 at 11 05 15@2x

The root bug was that we were always emitting fresh `implementationResult` vertexes because we didn't pass around `symbolInformationIDs` pointers. This explains why it always showed only one result for implementations.

olafurpg:
It's difficult to see the difference from the raw dump.lsif payloads. Here are screenshots showing the gist of the difference

CleanShot 2022-05-25 at 12 58 55
CleanShot 2022-05-25 at 12 57 44

PR #36, opened 2022-06-01
Fix a bug where goto definition did not work when SymbolInformation was missing

olafurpg:
### Test plan
See commit log for new snapshot tests.


cmd/tests/snapshots/output/missing-symbol-information/dump.lsif
```diff
@@ -0,0 +1,37 @@
+{"id":1,"version":"0.4.3","projectRoot":"file:/root","positionEncoding":"utf-8","toolInfo":{"name":"reprolang","version":"1.0.0"},"type":"vertex","label":"metaData"}
+{"id":2,"type":"vertex","label":"definitionResult"}
+{"id":3,"type":"vertex","label":"resultSet"}
+{"id":4,"type":"vertex","label":"referenceResult"}
+{"id":5,"type":"vertex","label":"hoverResult","result":{"contents":{"kind":"markdown"}}}
+{"id":6,"type":"edge","label":"textDocument/definition","inV":2,"outV":3}
+{"id":7,"type":"edge","label":"textDocument/references","inV":4,"outV":3}
+{"id":8,"type":"edge","label":"textDocument/hover","inV":5,"outV":3}
+{"id":9,"type":"vertex","label":"moniker","identifier":"reprolang repro_manager missing-symbol-information 1.0.0 globals.repro/NoSymbolInformation#","kind":"export","scheme":"reprolang"}
```
olafurpg, line 9: The bugfix was that this moniker is now `"kind":"export"` instead of `"kind":"import"`

cmd/tests/snapshots/output/missing-symbol-information/dump.lsif
```diff
@@ -0,0 +1,33 @@
+{"id":1,"version":"0.4.3","projectRoot":"file:/root","positionEncoding":"utf-8","toolInfo":{"name":"reprolang","version":"1.0.0"},"type":"vertex","label":"metaData"}
+{"id":2,"type":"vertex","label":"document","uri":"file:/root/globals.repro"}
+{"id":3,"type":"vertex","label":"range","start":{"line":0,"character":11},"end":{"line":0,"character":31}}
+{"id":4,"type":"vertex","label":"resultSet"}
+{"id":5,"type":"vertex","label":"referenceResult"}
+{"id":6,"type":"vertex","label":"hoverResult","result":{"contents":{"kind":"markdown"}}}
+{"id":7,"type":"edge","label":"textDocument/references","inV":5,"outV":4}
+{"id":8,"type":"edge","label":"textDocument/hover","inV":6,"outV":4}
+{"id":9,"type":"vertex","label":"moniker","identifier":"reprolang repro_manager missing-symbol-information 1.0.0 globals.repro/NoSymbolInformation#","kind":"import","scheme":"reprolang"}
```
olafurpg: Observe that the moniker has `"kind":"import"`

cmd/tests/reprolang/bindings/go/repro/scip.go
```diff
@@ -17,6 +19,11 @@ func (i *identifier) occurrence(roles scip.SymbolRole) *scip.Occurrence {
 func (s *reproSourceFile) symbols() []*scip.SymbolInformation {
 	var result []*scip.SymbolInformation
 	for _, def := range s.definitions {
+		fmt.Println("def.name.value", def.name.value)
+		if strings.Index(def.name.value, "NoSymbolInformation") >= 0 {
+			fmt.Println("BOOM")
```
varungandhi-src, line 24: Debugging line?

cmd/tests/reprolang/bindings/go/repro/scip.go
```diff
@@ -17,6 +19,11 @@ func (i *identifier) occurrence(roles scip.SymbolRole) *scip.Occurrence {
...
+			continue
+		}
```
varungandhi-src, line 26:
```suggestion
```

It seems like these lines are for debugging?

varungandhi-src (approved):
Looks OK modulo the stray `fmt.Println` lines.

cmd/tests/reprolang/bindings/go/repro/scip.go
olafurpg: Good catch! I opened the PR in a rush. Fixed now

tjdevries (approved):
Seems good, although I do wonder if we should add some warning/requirement/etc that tells you "hey, you have a non-local symbol that is a definition but we don't know anything about it". We could table that for now though, this is the right thing to do on the `scip` side of things

olafurpg:
> I do wonder if we should add some warning/requirement/etc that tells you "hey, you have a non-local symbol that is a definition but we don't know anything about it"

I think it's fine to leave out the `SymbolInformation` if it's not needed (no hover message, no relationships). It's best if we limit the number of rules that require runtime validation.

PR #37, opened 2022-06-01
Add a `Document.language` field

olafurpg:
Previously, SCIP consumers had to infer the programming language based
on the `Document.relative_path` field. It's common for some programming
languages to support multiple file formats (example, `.ts` and `.tsx`
for TypeScript).

Thank you @donsbot for the suggestion!

### Test plan
See the CI go green.


scip.proto
```diff
@@ -369,3 +375,88 @@ enum DiagnosticTag {
...
 }
+
+// Language standardises names of common programming languages that can be used
+// for the `Document.language` field. The primary purpose of this enum is to
+// prevent a situation where we have a single programming language ends up with
+// multiple string representations. For example, the C++ language uses the name
+// "C_PlusPlus" in this enum and other names such as "cpp" are incompatible.
+// Feel free to send a pull-request to add missing programming languages.
+enum Language {
+  UnspecifiedLanguage = 0;
```
varungandhi-src, line 386: Is there a specific reason these are kinda in haphazard order? It would be nice to have related languages be grouped to some extent.

scip.proto
```diff
@@ -369,3 +375,88 @@ enum DiagnosticTag {
...
+  Scala = 2;
+  Kotlin = 3;
+  Go = 4;
+  TypeScript = 5;
+  JavaScript = 6;
+  Python = 7;
+  Ruby = 8;
+  C_Sharp = 9;
+  C_PlusPlus = 10;
+  Objective_C = 11;
```
varungandhi-src, line 397: Any specific reason for mixing underscores instead of sticking to CamelCase? It seems a little odd; I don't think many languages follow this casing convention.

scip.proto
```diff
@@ -384,79 +384,93 @@ enum DiagnosticTag {
...
+  SQL = 69;
+  PLSQL = 70;
+  Prolog = 71;
+  Ini = 72;
+  TOML = 73;
+  YAML = 74;
+  JSON = 75;
+  Jsonnet = 76;
+  Nix = 77;
+  Bazel = 78;
```
olafurpg, line 464: Should this be Skylark? BUILD files and `.bzl` files are written in Skylark. There's no "Bazel" language to the best of my understanding

olafurpg (commented):
The order was mostly random. Ideally, we could make it alphabetical but that would make it complicated to manage the incrementing tag. What is the reasoning behind the current ordering?

olafurpg:
I updated the enum to use alphabetical ordering and renamed the languages to be consistent with LSP (example: "CPP" instead of "C_PlusPlus")

varungandhi-src (approved):

PR #38, opened 2022-06-01
schema: Update wording and punctuation

varungandhi-src:
While I was working on the documentation on how to write a SCIP indexer,
I ran into some ambiguities in the schema. So I've marked those with TODOs
in the last commit.

Other commits are minor cleanups and clarifications.

I'll regenerate the code after the other in-flight PR is merged and we've
settled on the wording here.

### Test plan

n/a -- only comments are updated.


scip.proto
```diff
@@ -240,25 +244,27 @@ enum SyntaxKind {
...
   // Function definition only
   IdentifierFunctionDefinition = 16;
 
-  // Macro call/reference
+  // Macro references, including invocations
   IdentifierMacro = 17;
   // Macro definition only
   IdentifierMacroDefinition = 18;
 
+  // TODO[Varun]: What does 'including namespaces' mean?
```
varungandhi-src, line 267: TODO for clarification.

scip.proto
```diff
@@ -326,18 +335,20 @@ message Occurrence {
   // (optional) The symbol that appears at this position. See
   // `SymbolInformation.symbol` for how to format symbols as strings.
   string symbol = 2;
-  // (optional) Bitmask for what `SymbolRole` apply to this occurrence. See
-  // `SymbolRole` for how to read and write this field.
+  // (optional) Bitset containing `SymbolRole`s in this occurrence.
+  // See `SymbolRole`'s documentation for how to read and write this field.
   int32 symbol_roles = 3;
-  // (optional) Markdown-formatted documentation for this specific range.  If
+  // TODO[Varun]: Specify dialect of Markdown.
```
varungandhi-src, line 341: TODO for discussion/clarification.

scip.proto
olafurpg: I'm guessing it's references to package identifiers, for example `java` and `io` in `java.io.File`. However, we should have a dedicated syntax kind for identifiers to packages/namespaces. IntelliJ gives them different colors than type identifiers and I find it very helpful to distinguish between them when reading code.

scip.proto
```diff
@@ -68,9 +68,13 @@ message ToolInfo {
...
-  // (Required) Path to the text document relative to the directory supplied in
-  // the associated `Metadata.project_root`. Not URI-encoded. This value should
-  // not begin with a directory separator.
+  // (Required) Unique path to the text document relative to the directory
+  // supplied in the associated `Metadata.project_root`. Not URI-encoded.
+  // This value should not begin with a directory separator.
+  //
+  // TODO: Specify if we allow both '/' and '\' as the path separator
+  // or if we expect Unix-style paths.
+  // TODO: Specify what should be done in the presence of symlinks.
```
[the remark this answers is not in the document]
olafurpg: In SemanticDB we use URI-encoding and maybe that's the best solution. It's simple to explain if we say that joining `Metadata.project_root` and `Document.relative_path` creates a URI to the file path. We should remove the "Not URI-encoded" comment above in that case.

Re symlinks: it's up to the indexer to decide. The consumer can figure out if it's a symlink or not so it's not a problem if `relative_path` points to a symlink.

scip.proto
olafurpg: I'd go with CommonMark rather ran GH-flavored https://commonmark.org

It seems like Prettier targeted CommonMark when they added support to format Markdown files https://prettier.io/blog/2017/11/07/1.8.0.html

olafurpg (commented):
These are good questions!

scip.proto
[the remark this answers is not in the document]
varungandhi-src: I'm not sure what you mean here in terms of URI encoding: I thought you can't have relative paths in URIs. Would the relative path look like `file:///dir/file`? Or do you mean that it would still be `dir/file`, which is then added as a suffix to the `Metadata.project_root`?

Wrt symlinks: the problem is one of merging `Document` structures. Let's say an `Index` contains a document with path `p` with some occurrences `po` and symbols `ps`. It also contains another document with path `q` with some occurrences `qo` and `qs`. Now if `q` is a symlink to `p`, what will Sourcegraph show when you navigate to `p` via `q`? The results from `qo` and `qs`? The results of merging `po`+`ps` and `qo`+`qs`? Just from `po`+`ps`? Should we flag anything here as a validation error/warning when uploading/ingesting the index?

scip.proto
varungandhi-src: We have `IdentifierModule`; that seems applicable here.

```
  // Identifier representing a unit of code abstraction:
  // `os` in `import os` in Python, `main` in `package main` in Go.
  IdentifierModule = 14;
```

It seems like we should reuse that for packages and namespaces, and remove the point about namespaces from this comment?

scip.proto
varungandhi-src: Removed TODO and replaced Markdown with CommonMark.

scip.proto
olafurpg: I missed that one. It's a bit confusing how we currently mix package/module/namespace in the spec. For example, we use `<package-name>` to refer to an npm package name, while `Descriptor.Suffix.Package` and `SyntaxKind.IdentifierModule` both refer to a C++ namespace, python/js module, or Java package.

olafurpg: It is BTW backwards compatible to rename `Suffix.Package` and `IdentifierModule` into something else. The term "namespace" might be the most self-explanatory name.

scip.proto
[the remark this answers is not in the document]
olafurpg: > Or do you mean that it would still be dir/file, which is then added as a suffix to the Metadata.project_root?

This one. We could require that the concatenation of `Metadata.project_root` and `Document.relative_path` makes a valid URI. This is what SemanticDB does but it's been problematic in my experience because very few (Java) libraries make it easy to get out the URI-encoded path part of a URI. In >>>99.9% of the time, it's not a problem because the URI encoding is the same as the non-URI encoding. In practice, I frequently didn't bother with URI-encoding and joined instead the individual path parts with a `/` separator.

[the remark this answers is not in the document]
olafurpg: I'd try something like this.

```proto
(Required) Relative path to the regular file on disk backing this text document.
Must point to a regular file, not a symbolic link.
Must be relative to the associated `Metadata.project_root`.
Must use forward slash separators "/" to join path parts, including on Windows.
Must be canonical, no parts of the path can be ".." and duplicate directory separators "//" are not allowed.
Must not begin with a forward slash "/".
Individual parts of the file path must not be URI-encoded.
```

The original version of SCIP had `Document.uri`, which I think is a lot simpler because you avoid having to document all of these requirements. We turned it into a relative path for the hypothetical use-case of simplifying incremental indexing, but I'm still not convinced this change was necessary for that use-case. The thinking was that the byte payload of a `Document` message should be consistent across machines if we remove absolute paths. 

The problem is that there's no standard for relative URI paths 🤷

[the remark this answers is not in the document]
varungandhi-src: Updated docs with requirements for paths.

scip.proto
varungandhi-src: We're consistently using namespace now, so this is resolved.

Issue #25, opened 2022-05-24
Replace docopt usage with urfave/cli


varungandhi-src:
It looks like https://cli.urfave.org/v2/ has a fair bit of functionality like completion etc, and Olaf mentioned that commands without `=` fail with cryptic errors.

```
$ scip convert --from index.scip --to dump.lsif

panic: interface conversion: interface {} is nil, not string

goroutine 1 [running]:
main.convertMain(0x1552c60?)
	/Users/olafurpg/dev/sourcegraph/scip/cmd/convert.go:15 +0x474
```

I suppose this error could be fixed by removing the `=` in the docopt text, but it's not intuitive that a space is not accepted, and the error message for the crash is also not helpful.

varungandhi-src:
Panic fixed in https://github.com/sourcegraph/scip/pull/26

Issue #19, opened 2022-05-23
Add process-level testing infrastructure


varungandhi-src:
As I [wrote earlier in a PR](https://github.com/sourcegraph/scip/pull/11#issuecomment-1130654590), we should add some process-level testing so that we can test the CLI more thoroughly.

This would involve:

- [ ] A small CLI for indexing reprolang code.
- [ ] Some tests for existing subcommands where we invoke the scip CLI with some arguments and examine the output/exit code and/or certain golden files.

PR #48, opened 2022-06-09
cli: Replace docopt with urfave/cli.

varungandhi-src:
Fixes issue #25.

### Test plan

Manually tested + updated existing tests. I'll add some follow-up tests as part of #19.


cmd/main.go
```diff
@@ -1,59 +1,60 @@
...
+				Destination: &versionFlag,
+				Aliases:     []string{"v"},
+			},
+		},
+		Action: func(c *cli.Context) error {
+			if versionFlag {
+				fmt.Println("0.1.0")
+				os.Exit(0)
+			}
+			// FIXME: What is the right way to print help text and error here?
```
varungandhi-src, line 44: I should file an issue to fix this if we don't know a good way to fix this right now.

olafurpg (approved):
LGTM 👍 I tried this locally and it's a great improvement

cmd/main.go
varungandhi-src: https://github.com/sourcegraph/scip/issues/49

PR #54, opened 2022-06-15
schema: Clarify module and package terminology.

varungandhi-src:
This lines up the terminology with RFC 698 and avoids confusion
due to usage of 'package' with two different meanings.

I renamed one of the packages to package-descriptor as a temporary
"fix" earlier, but I finally (re-)understood what it actually meant.

### Test plan

n/a


varungandhi-src:
This resolves part of the discussion in https://github.com/sourcegraph/scip/pull/38/files#r887249371

scip.proto
```diff
@@ -97,8 +97,8 @@ message Document {
 //   <manager>              ::= same as above, use the placeholder '.' to indicate an empty value
 //   <package-name>         ::= same as above
 //   <version>              ::= same as above
-//   <descriptor>           ::= <package-descriptor> | <type> | <term> | <method> | <type-parameter> | <parameter> | <meta>
-//   <package-descriptor>   ::= <name> '/'
+//   <descriptor>           ::= <module> | <type> | <term> | <method> | <type-parameter> | <parameter> | <meta>
```
olafurpg, line 100: What do you think about “namespace” instead of module?

scip.proto
```diff
@@ -127,8 +130,14 @@ message Package {
...
   enum Suffix {
+    option allow_alias = true;
     UnspecifiedSuffix = 0;
-    Package = 1;
+    // Unit of code abstraction and/or namespacing.
+    //
+    // NOTE: This corresponds to a package in Go and JVM languages.
+    Module = 1;
+    // Use Module instead.
+    Package = 1 [deprecated=true];
```
olafurpg, line 140: Nice. I didn’t know this was possible

scip.proto
```diff
@@ -127,8 +130,14 @@ message Package {
 
 message Descriptor {
   enum Suffix {
+    option allow_alias = true;
     UnspecifiedSuffix = 0;
-    Package = 1;
+    // Unit of code abstraction and/or namespacing.
+    //
+    // NOTE: This corresponds to a package in Go and JVM languages.
+    Module = 1;
```
olafurpg, line 138: Ditto: namespace

scip.proto
varungandhi-src: I think that's OK too. I've made that change, including a place which I missed earlier with `IdentifierModule`.

PR #58, opened 2022-06-21
Automatically add issues to Code Intel Project Board

olafurpg:
### Test plan

 n/a


varungandhi-src (approved):
If this is copied from somewhere else, it would be good to add a link.

olafurpg:
Good idea. There's no source of truth, but I linked to the scip-typescript copy that I used.

PR #66, opened 2022-07-12
schema: Document that names should be fully-qualified.

varungandhi-src:
I know it's kinda' "obvious" if you think about it for a bit, but I made
the mistake of following the words very literally and ended up having
a silly bug in scip-ruby with the first implementation. https://github.com/sourcegraph/scip-ruby/issues/14

### Test plan

n/a


scip.proto
```diff
@@ -119,6 +119,12 @@ message Document {
 //   <escaped-identifier>   ::= '`' { <escaped-character> } '`'
 //   <escaped-characters>   ::= any UTF-8 character, escape backticks with double backtick.
 // ```
+//
+// For "top-level" identifiers, such as types, fields/terms, methods, the
+// `<name>` should be a fully-qualified within the project. Generally, this
```
olafurpg, line 124: This sounds inaccurate, the `Symbol.descriptors` field (which maps to `{ <descriptor> }`) should be fully qualified while `Descriptor.name` should be a single identifier like the name of a method or field 🤔

varungandhi-src: Ah; it's the same confusion from before around the semantics of `{ }`. I'll replace that with a regex `+` and document that.

varungandhi-src: It seems like the namespace can/should contain the filepath though? Or do we not rely on that? That's not necessarily obvious from the text.

olafurpg: The descriptor doesn't need to include a file path, for JVM languages we only use packages (example `java/io/File#`). The only hard requirement is that the descriptor must be stable between compilations and match at the definition site and at reference site (which can be a separate repo from the definition). It's desirable, but optional, that the descriptor has human readable names.

varungandhi-src: So the complete list of descriptors has to be "fully qualified" when viewed as a whole, right?

In Java's case, IIUC, the path essentially ends up being encoded in the package structure anyways, so a separate path is not necessary.

olafurpg: > So the complete list of descriptors has to be "fully qualified" when viewed as a whole, right?

Correct.

varungandhi-src: Updated the wording; can you change a second look?

olafurpg (approved):
LGTM 👍 I like the new clarification, and `(PATTERN)+` syntax. The `{PATTERN}` syntax was borrowed from the Scala parser

PR #71, opened 2022-07-20
schema: Mention that non-local symbols don't need SymbolInformation.

varungandhi-src:
I was staring at some of the scip-ruby code I wrote earlier and I very literally interpreted the doc comment to emit `SymbolInformation` for all `Symbol`s, including local ones.

### Test plan

n/a


scip.proto
```diff
@@ -85,7 +85,7 @@ message Document {
   string relative_path = 1;
   // Occurrences that appear in this file.
   repeated Occurrence occurrences = 2;
-  // Symbols that are defined within this document.
+  // Non-local symbols that are defined within this document.
```
olafurpg, line 88: Where should local symbols have `SymbolInformation`? This comment somewhat implies that this list must contain only non-local symbols.

olafurpg: `SymbolInformation` should be optional for all symbols (local or not) after we merged this PR here https://github.com/sourcegraph/scip/pull/36

Before that PR, it was necessary to emit `SymbolInformation` for every symbol that has a definition occurrence.

varungandhi-src: I guess maybe I should reword it. The point I was trying to make was, that it isn't useful to have information about local symbols in this list (purely information in the occurrences list should be fine). Is that correct or am I misunderstanding the purpose of this list?

olafurpg:
Closing since this PR has gone stale and off-sync with the latest version. Feel free to re-open with a rebased diff on top of main.

PR #74, opened 2022-07-22
Rename `IdentifierKeyword` into `Keyword`

olafurpg:
I was surprised to see we have the name `IdentifierKeyword` because
keywords are not identifiers in the compiler/tokenizer APIs that I have
used in the past.

### Test plan

See the newly generated protobuf bindings.


varungandhi-src (approved):

tjdevries:
Sounds good to me. I think when we did this we were thinking more for syntax highlighting, so we were just adding identifier to a lot of stuff. :shipit:

PR #76, opened 2022-08-02
docs: Add use case for override_documentation.

varungandhi-src:
I'm going to need this in scip-ruby; I'm guessing scip-python uses this too?

### Test plan

n/a


tjdevries (approved):
LGTM

PR #77, opened 2022-08-02
docs: add some install instructions

ckipp01:
This pr just adds a few notes about installing the cli for those that
aren't as familiar with Go.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -39,6 +39,18 @@ to better understand project structure and common workflows.
 
 Contributors should abide by the [Sourcegraph Code of Conduct](https://handbook.sourcegraph.com/company-info-and-process/communication/code_of_conduct/).
 
+## Installing CLI
+
+You can find binaries for `scip`
+[here](https://github.com/sourcegraph/scip/releases). If you'd like to create a
+binary locally you can do so with:
+
+```sh
+git clone https://github.com/sourcegraph/scip.git --depth=1
+cd scip
+go build -o scip ./cmd 
+```
+
 ## SCIP CLI reference
 
 ```


Readme.md
```diff
@@ -39,6 +39,17 @@ to better understand project structure and common workflows.
 
 Contributors should abide by the [Sourcegraph Code of Conduct](https://handbook.sourcegraph.com/company-info-and-process/communication/code_of_conduct/).
 
+## Installing CLI
```
ckipp01, line 42: One thing to note is that with these instructions it installs a `cmd` as the executable name. I'm assuming that's not desired, but I'm unsure if that's just due to the project structure, a bug, or I'm not installing it right.

varungandhi-src: Normally, we'd provide binaries which are named correctly, but we haven't gotten around to doing that. We should fix that.

For building from source, this works:

```
git clone https://github.com/sourcegraph/scip.git --depth=1
cd scip
go build -o scip ./cmd 
```

`go install` does not support renaming the binary. https://github.com/golang/go/issues/44469 So a `mv` is needed.

The `cmd` naming is from the Go naming convention around CLIs. I'm not a fan of it but that is what it is. 🤷🏽 

As for `GOPATH`, that's not strictly needed. Instead, you can do `GOBIN=SOME_PATH go install ...`. So long as `SOME_PATH` is part of `PATH`, it should be fine.

Readme.md
```diff
@@ -39,6 +39,17 @@ to better understand project structure and common workflows.
...
+
+You can install the `scip` cli by:
```
varungandhi-src, line 44:
```suggestion
You can install the `scip` CLI by:
```

varungandhi-src:
Apologies for missing this earlier. If you're submitting a PR to one of the repos, feel free to @ one of us directly if your PR hasn't been reviewed in a few days.

Readme.md
ckipp01: > Normally, we'd provide binaries which are named correctly, but we haven't gotten around to doing that. We should fix that.

This would be ideal.

I rephrased it to just simplify it and show the approach you outlined.

varungandhi-src: I've pushed a change to provide binaries and created a new tag. Hopefully, that generation + upload should happen here. https://github.com/sourcegraph/scip/actions/runs/2819700747

varungandhi-src: Could you update this to recommend downloading a release first? We have release binaries up now: https://github.com/sourcegraph/scip/releases/tag/v0.2.0

varungandhi-src (approved):
Approving modulo the minor suggestion to recommend downloading a release binary.

Readme.md
ckipp01: Sure, updated. Thanks for publishing the binaries!

PR #82, opened 2022-08-17
feat: add folding ranges

tjdevries:
### Test plan
- [ ] ya


olafurpg:
As discussed offline, let's keep folding ranged outside of SCIP because folding ranges can be implemented with a syntactic tool like tree-sitter and doesn't benefit from semantic understanding of the code. One risk of using entirely different folding range implementations for different languages is that the folding behavior becomes inconsistent. I've personally never gotten much utility out of folding ranges in VS Code (partly because of how inconsistently it's implemented by different languages), compared to IntelliJ where I use folding ranges very frequently (although partly because the interaction between the cursor and folded ranges is significantly better in IntelliJ compared to VS Code).  It's desirable to keep SCIP small and adding code folding ranges would be an unnecessary addition for now, although we can always reconsider this decision in the future if there's strong demand to include folding ranges in the protocol.

PR #86, opened 2022-09-19
Add github repository link

est31:
Repository links are useful if you want to figure out where a particular crate is hosted. I have wondered the same thing and was able to find this repository, but it would have been easier if I could just have clicked the link on docs.rs.


varungandhi-src:
Thanks!

PR #88, opened 2022-09-22
WIP: Add is_definition.

varungandhi-src:
- WIP: Add is_definition.
- test: Add separate relationships statement to repro.

### Test plan

Going to see if the code navigation works as expected in an SG instance with the new relationships statement in repro, which allows adding a relation to a symbol which lacks a definition.


varungandhi-src:
Looks like this does have the desired effect. 🎉

olafurpg (approved):
LGTM. 

I see the hover is empty, should we add is_hover as well?

varungandhi-src:
IIUC, the hover is empty because I set the documentation to be blank. So I don't quite get what `is_hover` would solve... If A is a definition for B, isn't the desired behavior that you show hover docs of A when you hover over B?

varungandhi-src:
Let's merge this for now so that I can start using it in scip-ruby. We can add is_hover later too.

PR #92, opened 2022-09-30
cli: Add lint subcommand.

### LINKED ISSUE #90: Optionally diagnose ill-formed-ness related issues with indexes
Rough sketch of what this would look like:
- Take a flag describing whether diagnostics should be emitted or not (or have a separate function for this which can be optionally called)
- Once the SCIP index has been ingested, perform analyses about well-formedness if the flag was passed
- Print diagnostics in `scip convert` (not suppressed by default) and in `src-cli` (suppressed by default)
- Optional: Add new `scip diagnose` subcommand. If we add this, then maybe `scip convert` doesn't need to warn by default.

varungandhi-src:
Fixes #90.

### Test plan

Tested manually against an index generated by scip-ruby, and it did flag a bunch of errors.

TODO:
- [x] Add some automated tests for the different error paths.

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -50,6 +50,14 @@ and can be inspected using `protoc`:
 protoc --decode=scip.Index -I /path/to/scip scip.proto < index.scip
 ```
 
+There is also a `lint` subcommand which performs various well-formedness
+checks on a SCIP index. It is meant primarily for people working on a SCIP indexer,
+and is not recommended for use in other settings.
+
+```
+scip lint /path/to/index.scip
+```
+
 ## Testing and adding new SCIP semantics
 
 It is helpful to use reprolang to check the existing code navigation behavior,
--- a/Readme.md
+++ b/Readme.md
@@ -70,9 +70,10 @@ DESCRIPTION:
 
 COMMANDS:
    convert   Convert a SCIP index to an LSIF index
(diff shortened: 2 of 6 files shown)


olafurpg (approved):
LGTM 👍🏻 I didn't closely review the details of the linting logic itself but I'm happy to merge this as is.

Some concerns I have with this linting system:

- There's no way to disable certain groups of issues
- There's no distinction between warnings and errors
- Some errors presumably have location information (path:line) but it doesn't look like the errors are formatted that way
- There's no way to test the linting logic if we want to add new linting rules

None of those are blocking issues.

varungandhi-src:
> There's no way to disable certain groups of issues

This can be added later. One can use `grep -v` in the meantime.

> There's no distinction between warnings and errors

Do you mean in terms of formatting, in terms of the exit code, or something else?

> Some errors presumably have location information (path:line) but it doesn't look like the errors are formatted that way

I can modify the format for occurrence related errors to tweak this.

> There's no way to test the linting logic if we want to add new linting rules

I will add that before merging -- the PR is currently WIP (I didn't mark it as draft because implementation stuff is mostly done, but testing is remaining).

varungandhi-src:
I've added test cases for all the warnings and errors. We can tweak the error format later as needed.

PR #95, opened 2022-10-01
fix: view -> print in docs

ckipp01:
### Test plan

All tests should just continue to pass.


varungandhi-src:
Thanks, forgot to update this, oops.

PR #103, opened 2022-10-06
cli: Improve version reporting.

### LINKED ISSUE #99: Better automation for --version
Idea:
- Use a separate `cli_version.txt` file (or similar).
- Embed the file inside the binary and use that for printing the version.
- Create a small release shell script which adds both the git tag, modifies the version file and does anything else that is necessary. 
- Add a check in the release pipeline that for a release binary, the compiled binary's printed `--version` is the same as the tag. (To avoid an accidental trigger of the release pipeline due to a direct tag push to actually publish a release with the wrong `--version`).

varungandhi-src:
Fixes #99.

### Test plan

Manually tested version output. It looks like:

```
scip version v0.2.1-dev
SHA: fdda0abfa1fa46af2fa12b39dba50c6f4bbd7649
timestamp: 2022-10-06T03:50:35Z
clean: false
```

in dev builds, and should look like `scip version v0.2.1` in CI builds.


olafurpg (approved):

cmd/main.go
```diff
@@ -23,10 +27,38 @@ func commands() []*cli.Command {
 	return []*cli.Command{&convert, &lint, &print, &snapshot, &stats}
 }
 
+//go:embed version.txt
```
abitrolly, line 30: Nice. Didn't know that is possible.

PR #108, opened 2022-10-07
bindings/go: Fix conversion for is_definition relationships.

varungandhi-src:
### Test plan

Manually tested against the index generated by scip-ruby.


bindings/go/scip/convert.go
```diff
@@ -286,19 +299,14 @@ func (g *graph) emitRelationship(relationship *Relationship, rangeID, documentID
 			// The 'property' field is included in the LSIF JSON but it's not present in reader.Element
 			// Property: "referenceResults",
 		})
-		return []int{relationshipIDs.ReferenceResult}
+		out = append(out, relationshipIDs.ReferenceResult)
```
olafurpg, line 302: I guess this diff is no longer needed, but it's good to have the accumulated logic in case we add another condition in the future

olafurpg (approved):
LGTM

bindings/go/scip/convert.go
varungandhi-src: Yep, I kept it to avoid the future bug.

PR #112, opened 2022-10-25
Re-pin TypeScript protoc generator.

olafurpg:
The latest build was failing for me locally, I'm not sure how it passed CI.

### Test plan
See the CI go green, and manually verify the job actually builds the TS part


varungandhi-src:
> I'm not sure how it passed CI.

Hmm, this is weird, I can take a look at this later, we should probably fix this sooner rather than later.

varungandhi-src (approved):

olafurpg:
I pushed a commit that should have failed the typecheck job but it succeeded

olafurpg:
Opened a followup issue https://github.com/sourcegraph/scip/issues/113

PR #118, opened 2022-11-16
Snapshot: don't fail on unparseable symbols.

olafurpg:
Previously, the `scip snapshot` command failed if there was a symbol that wasn't parseable. This was unhelpful when trying to debug why navigation was not working as expected (which is the main situation where I use `scip snapshot`). Unparseable symbols are not invalid, they just don't work with cross-repo navigation. To fail fast like before, there's a new `scip snapshot --strict` flag that fails if there's a parser error.

### Test plan

Manually tested locally.

### DIFF
--- a/bindings/go/scip/symbol_formatter.go
+++ b/bindings/go/scip/symbol_formatter.go
@@ -8,6 +8,7 @@ import (
 // Excluding parts of the symbol can be helpful for testing purposes. For example, snapshot tests may hardcode
 // the package version number so it's easier to read the snapshot tests if the version is excluded.
 type SymbolFormatter struct {
+	OnError               func(err error) error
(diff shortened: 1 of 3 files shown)


varungandhi-src:
Perhaps turn the error into a warning instead of dropping altogether? Since this is meant for dev use (vs an end-user thing), I think it'd be better to surface well-formedness issues by default, and have a way to turn them off, rather than the other way around.

olafurpg:
I enabled strict by default and updated the error message to explain how to disable strict. I agree it's desirable to surface these errors by default, thank you raising that point. I'm not a fan of warnings, they tend to become noise.

varungandhi-src (approved):

PR #44, opened 2022-06-08
add symbol support and move generated code for rust

tjdevries:
Improve rust bindings and provide symbol library like we have for Golang.

### Test plan

- [ ] TODO

PR #129, opened 2023-03-09
docs: Document suffix for macros

varungandhi-src:
The suffix '!' was added in https://github.com/sourcegraph/scip/pull/44
but the documentation was not updated.

### Test plan

n/a


tjdevries (approved):
Ah, good catch. thanks

PR #130, opened 2023-03-13
Remove debugging output from symbol.rs that cannot otherwise be disabled.

asutherland:
As documented at https://doc.rust-lang.org/std/macro.dbg.html the dbg! macro unconditionally prints its argument to stderr, even in release builds.  (It is not possible to quiet it through use of RUST_LOG.)

Presumably this was accidentally left in.


tjdevries:
Yup! sorry about that (and the delay). Appreciate the fix

PR #132, opened 2023-03-18
[rust bindings] Allow zero-length method disambiguation.

asutherland:
parse_symbol was failing on a bunch of rust symbols that look like "rust-analyzer cargo test_rust_dependency 0.1.0 MyType#new()." with `Err(InvalidIdentifier("method disambiguator"))`.

The problem is that the method descriptor was attempting to use peek to look for the character after the opening "(" to see if it was a ")", but the index had already been incremented, so peek would see the "." closing character instead of the ")".

I also have a change in this patch to disable automatic doctests because `cargo test` is failing for me locally without this because it seems to view the inlined protobuf schema as a doctest and reasonably has trouble parsing the schema as valid rust code.  I am not able to delve into the general machinery around the "generated" directory at this time, but it seems nice for the tests to pass, so this seems like a reasonable band-aid.

### DIFF
--- a/bindings/rust/Cargo.lock
+++ b/bindings/rust/Cargo.lock
@@ -65,9 +65,9 @@ dependencies = [
 
 [[package]]
 name = "protobuf"
-version = "3.1.0"
+version = "3.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


tjdevries:
Awesome :) Thanks! This was a good catch.

Issue #111, opened 2022-10-11
snapshot: use correct language-specific comment token for snapshot output


Strum355:
Noticed when trying out scip-ruby that the snapshot output uses `//` regardless of language, which is the wrong comment token for ruby. This used to be a thing in the old snapshot code in scip-java I believe, so itd be cool to reintroduce it here.

varungandhi-src:
Yeah, I was going to make it customizable, but we didn't change it because this is mostly used for debugging. The snapshot code in scip-ruby uses `#`, and it is ported from the code here but tailored a bit so that we can evolve it independently (and avoids a dependency on a Go project from a C++ project).

PR #139, opened 2023-04-03
feat: add comment-syntax flag

tjdevries:
Fixes: #111

### Test plan

- [ ] Use on something like a python project to do `# ` as the prefix, instead of `//`

### DIFF
--- a/cmd/snapshot.go
+++ b/cmd/snapshot.go
@@ -13,9 +13,10 @@ import (
 )
 
 type snapshotFlags struct {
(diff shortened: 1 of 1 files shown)


Strum355 (approved):
Approving, cant check from mobile but does this properly handle comment syntaxes that are different in length than //? I feel like there could be a case of one-off if using # but that might already be handled

tjdevries:
It checks for the length of the comment syntax provided.

PR #141, opened 2023-04-03
ci: add running rust tests

tjdevries:
Enable running rust tests in CI

### Test plan
- [ ] Green tests (now actually running cargo test)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,6 +17,13 @@ jobs:
         with:
           rust: true
       - run: cargo check
-        working-directory: bindings/rust
-      - run: cargo check
         working-directory: cmd/tests/reprolang
+  rust-bindings-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/asdf
+        with:
+          rust: true
+      - run: cargo test
+        working-directory: bindings/rust


.github/workflows/rust.yml
```diff
@@ -16,7 +16,42 @@ jobs:
...
+      - name: Cache cargo target
+        uses: actions/cache@v2
+        with:
+          path: target
+          key: ${{ runner.os }}-cargo-target-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-target-
       - run: cargo check
         working-directory: cmd/tests/reprolang
+  rust-bindings-test:
```
varungandhi-src, line 35: Could we directly use https://github.com/Swatinem/rust-cache instead of having this custom code with key-specificity etc here?

.github/workflows/protobuf.yml
```diff
@@ -19,5 +19,11 @@ jobs:
           js: true
           rust: true
           golang: true
+      - name: Cache `.bin` folder
+        id: cache-setup
+        uses: actions/cache@v2
+        with:
+          path: .bin
+          key: ${{ runner.os }}-bin
```
varungandhi-src, line 27: This seems insufficient, as the key won't be invalidated when the `dev/proto-generate.sh` file changes (as one example).

PR #142, opened 2023-04-11
Add --project-root option to override the root inside scip files

keynmol:
This option can be useful if the scip file is produced inside a docker container, and the project root in the file points to a non-existent location.

### Test plan

Manual testing

### DIFF
--- a/bindings/go/scip/testutil/format.go
+++ b/bindings/go/scip/testutil/format.go
@@ -19,7 +19,7 @@ func FormatSnapshots(
 	index *scip.Index,
 	commentSyntax string,
 	symbolFormatter scip.SymbolFormatter,
-	root string,
+	customProjectRoot string,
 ) ([]*scip.SourceFile, error) {
 	var result []*scip.SourceFile
 	projectRootUrl, err := url.Parse(index.Metadata.ProjectRoot)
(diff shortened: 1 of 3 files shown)


cmd/snapshot.go
```diff
@@ -36,6 +37,11 @@ and symbol information.`,
 				Destination: &snapshotFlags.output,
 				Value:       "scip-snapshot",
 			},
+			&cli.StringFlag{
+				Name:        "project-root",
+				Usage:       "Override projet root in SCIP file",
```
olafurpg, line 42:
```suggestion
				Usage:       "Override project root in the SCIP file. For example, this can be helpful when the SCIP index was created inside a Docker image or created on another computer.",
```

bindings/go/scip/testutil/format.go
```diff
@@ -18,25 +19,36 @@ func FormatSnapshots(
 	index *scip.Index,
 	commentSyntax string,
 	symbolFormatter scip.SymbolFormatter,
+	root string,
```
olafurpg, line 22: ?
```suggestion
	customProjectRoot string,
```

olafurpg (changes requested):

olafurpg (approved):
Feel free to merge after addressing the comments. This is a great addition!

PR #147, opened 2023-05-02
Allow printing SCIP index as JSON

keynmol:
Goals:

1. ✅  Adds a `--json` flag to `print` command
3. ✅  Updates CLI docs as they've grown out of date~

Stretch goals:
- ~Fix (?) the testing commands as it seems they're not running on CI (i.e. CLI docs are out of date)~ Done in #148 

### Test plan

- [x] Add basic Go tests


cmd/json.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func jsonCommand() cli.Command {
+	var jsonFlags jsonFlags
+
+	json := cli.Command{
+		Name:  "json",
+		Usage: "Print SCIP index as JSON",
+		Flags: []cli.Flag{
+			fromFlag(&jsonFlags.from),
```
keynmol, line 24: If anyone knows how to make this argument positional, I'd appreciate it
Just using the NArgs leads to the flags being ignored

`scip json -p --from index.scip` is not terrible, but leaves a lot to be desired

docs/CLI.md
```diff
@@ -96,8 +108,12 @@ DESCRIPTION:
...
 
 OPTIONS:
-   --from value  Path to SCIP index file (default: index.scip)
-   --to value    Path to output directory for snapshot files (default: scip-snapshot)
+   --comment-syntax value  Comment syntax to use for snapshot files (default: "//")
+   --from value            Path to SCIP index file (default: "index.scip")
+   --help, -h              show help (default: false)
+   --project-root value    Override project root in the SCIP file. For example, this can be helpful when the SCIP index was created inside a Docker image or created on another computer
+   --strict                If true, fail fast on errors (default: true)
+   --to value              Path to output directory for snapshot files (default: "scip-snapshot")
```
varungandhi-src, line 116: Q: Why is this PR changing these lines? I thought the earlier PR that verified the documentation should've fixed this... 🤔

keynmol: The tests only verify that the global CLI help text is valid, not for any of the subcommands :(

It won't be much of an issue to add extra checks in a separate PR.

docs/CLI.md
```diff
@@ -1,5 +1,16 @@
...
 
+<!--toc:start-->
+
+- [SCIP CLI Reference](#scip-cli-reference)
+  - [`scip convert`](#scip-convert)
+  - [`scip lint`](#scip-lint)
+  - [`scip print`](#scip-print)
+  - [`scip snapshot`](#scip-snapshot)
+  - [`scip stats`](#scip-stats)
+  - [`scip json`](#scip-json)
```
varungandhi-src, line 11: Nit: The `scip json` subcommand breaks alphabetical ordering.

docs/CLI.md
varungandhi-src: Could you remove the unrelated changes from this PR? I can submit a PR tomorrow making the test more robust and fix this.

keynmol:
> Subcommands should ideally be verbs, not nouns. That's already the pattern with existing subcommands (thought not all of them follow this). Thoughts on making this an extra --format= flag to scip print instead?

I think an even better option would be to add a `--json` flag to `print` subcommand, as it's a more intuitive and recommended way of controling output for CLI https://clig.dev/#output. **TODO: remove subcommand, make it a `--json` flag** 

> What's the motivation behind this subcommand? Is the output meant to be "stable" for some definition of stable? Since it's JSON, it's not unreasonable that someone may try to consume it. (Not saying it's a bad idea, but the motivation should at least be documented somewhere in the code/commit message etc.)

The motivation wouls be to provide a machine readable and ubiqutous output format, something amenable to other CLI tools like `jq. I looked around and there doesnt' seem to be an obvious way to inspect protobuf files in a dynamic fashion, like what `jq` provides. And even if there is such a tool, it's not as ubiqutous as JSON anyways.

w.r.t. stability, bar some variations in how to output missing fields, my understanding is that protobuf -> json conversion is unversal? So we can advertise it as stable. 

> Thoughts on skipping the pretty formatting? One can easily pipe the result through jq if pretty formatting is desired.

Yeah, I agree here. Once this becomes `--json` flag on the `print` command, there won't be a meaningful way to control the output, so we can just assume people will use `jq`. **TODO: remove pretty-printing**

docs/CLI.md
keynmol: > Could you remove the unrelated changes from this PR?

IMO, this is not a codebase with ridiculous velocity where unrelated changes are a problem - ongoing small improvements that aren't at odds with main purpose of the PR shouldn't be penalised by extra effort of reverting and resubmitting. 

There are many small things that can be improved for which there is no plan and no dedicated effort - with a small number of contributors we should aim to fix things as we go.

You are free to follow up with a test for this (or I can do this) but updating and fixing small things in the docs should be a welcome addition to any PR.

docs/CLI.md
keynmol: Will be removed entirely once the command becomes a flag instead

keynmol:
urfave/cli doesn't support positional arguments (https://github.com/urfave/cli/issues/1074)

So the only form in which the command works is `scip print --json <file>`, not `scip print <file> --json`

It's annoying, but an alternative would be to manually process the arguments, which IMO isn't worth it at this point.

docs/CLI.md
varungandhi-src: First of all, my comment was not a blocking one.

> ongoing small improvements that aren't at odds with main purpose of the PR shouldn't be penalised by extra effort of reverting and resubmitting.

I'm not trying to "penalise" something here... Ideally, these would not be part of the same PR (or commit) anyways. Normal commit hygiene IMO dictates that you'd commit related stuff in a single commit and move unrelated commit into separate PRs before submitting PRs. Even it got bundled into the same PR, if it were in different commits, it'd be easy to revert and push. 

That's regardless of codebase velocity etc. My point isn't about velocity, it's about completeness. With drive-by improvements that get mixed in with other PRs, it's less clear what still needs fixing vs what doesn't. E.g. here, if I hadn't flagged this, then the point about the test missing might not have been noticed.

That said, I don't think it's a big deal to leave it in.

varungandhi-src (approved):
It seems a little weird to me the CLI guidelines recommend separate `--json` and `--plain` flags rather than a single `--format=(plain|json)` flag, but it's not a big deal.

PR #150, opened 2023-05-02
Add `enclosing_range` to `Occurrence` message

efritz:
This PR adds a field to `Occurrence` which we plan to populate as a by-product of symbol generation or highlighting. We will consume this almost immediately in the Cody context APIs to break off only the relevant portion of a code intelligence result. This will also allow us to generate more intelligent embeddings chunks. This can also help us determine call hierarchies as we now have a way to extract symbols within the implementation of a given symbol (by range).

### Test plan

N/A


scip.proto
```diff
@@ -383,10 +383,31 @@ message Occurrence {
...
+  // func Parse(input string) string {
+  //     return input[n:]             
+  // }                                
+  // ```
+  // This field can also be used, for further example, to indicate the relevant parent 
+  // expression a variable referencing an invokable, e.g.:
+  // ```
+  // let x = a.b().f().g()
+  //               ^ range
+  //         ^^^^^^^ enclosing_range
```
olafurpg, line 408: My expectation is that the enclosing range would include the function arguments as well
```suggestion
  //         ^^^^^^^^^ enclosing_range
```

scip.proto
```diff
@@ -383,10 +383,31 @@ message Occurrence {
...
   // Benchmarks revealed that this encoding was inefficient and that we could
   // reduce the total payload size of an index by 50% by using `repeated int32`
-  // instead.  The `repeated int32` encoding is admittedly more embarrassing to
+  // instead. The `repeated int32` encoding is admittedly more embarrassing to
   // work with in some programming languages but we hope the performance
   // improvements make up for it.
   repeated int32 range = 1;
+  // (optional) The source position of the nearest non-trivial enclosing AST node.
+  // If supplied, the value must mirror the constraints detailed in the `range` field.
+  // For occurrences referencing the identifier within a class or function declaration,
```
olafurpg, line 392: I usually say "definition occurrence" to refer to occurrences that have the definition role
```suggestion
  // For definition occurrences, the enclosing range should encode the start/end bounds of the entire definition AST node.
```

scip.proto
```diff
@@ -383,10 +383,31 @@ message Occurrence {
...
+  // }                                
+  // ```
+  // This field can also be used, for further example, to indicate the relevant parent 
+  // expression a variable referencing an invokable, e.g.:
+  // ```
+  // let x = a.b().f().g()
+  //               ^ range
+  //         ^^^^^^^ enclosing_range
+  // ```
+  repeated int32 enclosing_range = 7;
```
olafurpg, line 410: Would be good to list example applications that could use enclosing_range:

- Call hierarchies
- Symbol outline (breadcrumbs in the editor)
- "Expand selection"
- Highlight range on hover

olafurpg (changes requested):

olafurpg (approved):
I took the liberty to push a commit that addresses my comments. LGTM 👍🏻

olafurpg:
A good followup PR would be to add `enclosing_range` to the snapshot output so that we can more easily test it. It's not a blocking issue since we will most likely need to rewrite the snapshot format in Rust for syntax-highlighter.

PR #151, opened 2023-05-03
Update buf.gen.yaml file

mfridman:
Hey there, this PR updates the plugin reference in buf.plugin.yaml. We have some [upcoming Remote Package changes](https://buf.build/blog/remote-packages-remote-plugins-approaching-v1) and would like to avoid breaking upstream OSS projects.

Note, I bumped the plugin version to the latest available and generated the `*.pb.go` files with that updated plugin.


buf.gen.yaml
```diff
@@ -6,7 +6,7 @@ plugins:
     out: ./docs/
     opt: docs/scip.sprig,scip.md
     # NOTE: the sprig file is a template, it is not generated.
-  - remote: buf.build/library/plugins/go:v1.27.1-1
+  - plugin: buf.build/protocolbuffers/go:v1.30.0
```
mfridman, line 9: Newer versions of the `buf` CLI use the `plugin` key for both local and remote plugin references.

varungandhi-src:
Thanks for the PR.

I tried this, the problem is that the generated doc comments with the version bump don't have the right number of spaces. 😐 Part of the doc comments which indicate source ranges become incorrect due to this...

I've created a separate PR to iterate. https://github.com/sourcegraph/scip/pull/157

Issue #154, opened 2023-05-15
Proposal: add several fields to `SymbolInformation`


olafurpg:
This issue is an umbrella for several proposed additions to SCIP, based on discussions with @donsbot [on Mastodon](https://aus.social/@donsbot/110343646389564312) .

## `SymbolInformation.display_name`

This would be the name of the symbol, which is both helpful for local variables and avoids parsing the name from the symbol. The field could be `name` instead of `display_name`, we use `display_name` in SemanticDB to emphasize that this name is meant to be displayed (and should therefore not have special encoding for non-ASCII characters like emojis.

## `SymbolInformation.owner`

Alternative name `parent`. The thinking with this field is that it avoids parsing the owner from the symbol, and it allows us to emit an owner for local symbols.

## `SymbolInformation.kind`

An enum that specifies what kind of symbol this is (enum/interface/method/...). Currently, `Descriptor.Suffix` doesn't encode enough fine-grained information (and it's intentionally named "suffix" to emphasize that it's primarily related to the syntax of the symbol.

## `SymbolInformation.signature_documentation`

A string-formatted rendering of the signature. Currently, indexers emit this information in the `documentation` field as markdown-formatted code blocks. Having a separate field makes it cleaner to extract only the signature. I propose we reserve the field `SymbolInformation.signature` for fully typed/structured signatures (not string-formatted signatures).

donsbot:
# SymbolInformation.kind

For symbol kinds, a variant of the lsif/vscode spec. E.g. currently used by Glean : https://github.com/facebookincubator/Glean/blob/main/glean/schema/source/codemarkup.types.angle#L38 ;  these are used for search and outline views: to e.g. constraint a search to methods or classes only, or when searching by name, to show the kind of each result.

# SymbolInformation.signature_documentation

this is somewhat important, as the scip index conflates "Documentation". e.g. in rust-analyzer the type signature and the documentation markdown are combined. They should ideally be separate "tables" in the output, keyed by scip.Symbol, as depending on the scenario you might only want to show the signature (e.g. a search page), or docs (an API page).

olafurpg:
@donsbot Agreed! We have these pieces of information in SemanticDB for similar reasons. If we add these fields to SCIP, would be interested in contributing the changes to rust-analyzer? I estimate it will be easy to update scip-java to emit the new fields. We can create a tracking issue for the remaining languages.

donsbot:
[...]

PR #156, opened 2023-05-23
Add `SymbolInformation.kind`

olafurpg:
This is the first step towards closing https://github.com/sourcegraph/scip/issues/154.

Previously, the only way to determine the type of a symbol was through `SymbolDescriptor.Suffix`. This solution was not ideal since `Suffix` had very coarse grained types, and it was intentionally named "suffix" to indicate it was primarily used to determine the syntax of the symbol (not semantics). 

This PR adds a new enum `SymbolInformation.Kind` to add first-class support to determine the type of a symbol.

### Test plan

Green CI.


olafurpg:
cc/ @donsbot if you have a moment to review.

scip.proto
```diff
@@ -184,8 +184,59 @@ message SymbolInformation {
...
+  // the same descriptor: `SymbolDescriptor.Suffix.Type`.
+  enum Kind {
+      UnspecifiedKind = 0;
+      Array = 1;
+      Boolean = 2;
+      Class = 3;
+      Constant = 4;
+      Constructor = 5;
+      Enum = 6;
+      Enumerator = 7;
```
varungandhi-src, line 206: Q: What is the difference between `Enum` and `Enumerator`?

scip.proto
```diff
@@ -184,8 +184,59 @@ message SymbolInformation {
   repeated string documentation = 3;
   // (optional) Relationships to other symbols (e.g., implements, type definition).
   repeated Relationship relationships = 4;
+  // The kind of this symbol. Use this field instead of
+  // `SymbolDescriptor.Suffix` to determine whether something is, for example, a
+  // class or a method.
+  Kind kind = 5;
+  // Kind determines the type of a symbol using idiomatic names from the
+  // language's type system. For example, a Java method would have the kind
```
varungandhi-src, line 192: This wording is a bit confusing since the "kind" is not the usual "kind" in type system jargon, "type" used as "category" again not in the sense of type systems, and yet there is a mention of a type system at the end of the sentence.

I would change it something like:

> // Kind represents the fine-grained category of a symbol, suitable for presenting
> // information about the symbol's meaning in the language.

(That's a bit too vague perhaps, but I would avoid using the words "type" and "type system" here.)

Maybe there should be an extra: "Since Kind is more fine-grained than Suffix, if two symbols have the same Kind, they should share the same Suffix. Conversely, if two symbols have different Suffixes, they should have different Kinds".

scip.proto
```diff
@@ -184,8 +184,59 @@ message SymbolInformation {
...
+      Namespace = 18;
+      Null = 19;
+      Number = 20;
+      Object = 21;
+      Operator = 22;
+      Package = 23;
+      PackageObject = 24;
+      Parameter = 25;
+      Property = 26;
+      SelfParameter = 27;
```
varungandhi-src, line 226: Suggestion: Add a `ThisParameter` too

scip.proto
```diff
@@ -184,8 +184,59 @@ message SymbolInformation {
...
+      String = 28;
+      Struct = 29;
+      Trait = 30;
+      Type = 31;
+      TypeParameter = 32;
+      Union = 33;
+      Value = 34;
+      Variable = 35;
```
varungandhi-src, line 234: Wondering if we should split this into `LocalVariable` and `GlobalVariable`?

scip.proto
```diff
@@ -184,8 +184,59 @@ message SymbolInformation {
...
+      Method = 16;
+      Module = 17;
+      Namespace = 18;
+      Null = 19;
+      Number = 20;
+      Object = 21;
+      Operator = 22;
+      Package = 23;
+      PackageObject = 24;
+      Parameter = 25;
```
varungandhi-src, line 224: Suggestion: Maybe add a `ParameterLabel` too (although, Swift uses the term "argument label" instead 🤷🏽)

scip.proto
```diff
@@ -184,8 +184,59 @@ message SymbolInformation {
...
+      Property = 26;
+      SelfParameter = 27;
+      String = 28;
+      Struct = 29;
+      Trait = 30;
+      Type = 31;
+      TypeParameter = 32;
```
varungandhi-src, line 231: Suggestion: Add `TypeClass`, `TypeFamily`, `DataFamily`, `Protocol`.

varungandhi-src (approved):

scip.proto
donsbot: enum T = { A , B , C }. T = the "Enum", A, B , C are "Enumerators" (in some languages, e.g. C++ or PHP)

scip.proto
donsbot: See also the LSP spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolKind

olafurpg: Good question. In SemanticDB with a `LOCAL` kind. My thinking was it's desirable to match LSP and it's possible to distinguish between local and global variables based on the symbol.

scip.proto
olafurpg: Are these for a specific language? I think we should be fairly liberal about adding language-specific kinds so we don't have to capture everything in this PR (we just need to make sure that there are no unintended conflicts among the kinds we're introducing, like `EnumMember` and `Enumerators`).

scip.proto
olafurpg: Sounds good to me. For which language is that?

scip.proto
olafurpg: > I would avoid using the words "type" and "type system" here

Agreed. The docstring here was written sloppily.

I like your proposed reformulation, and I think it's a good idea to mention the expected relationships between suffix and kinds.

scip.proto
varungandhi-src: C++, JavaScript and Java use `this` whereas Python, Rust and Swift use `self`.

olafurpg: `SelfParameter` is actually used to give a name for `this` in Scala

```
class Foobar { self =>
  val a = 1
  class Inner {
    val a = 2
    def add = a + self.a (
  }
}
```

varungandhi-src:
cc @donsbot @olafurpg in case you have any more feedback. I may have gone a bit over-the-top with adding kinds of many different languages. 😅 

If there's no more feedback, we can merge this.

PR #158, opened 2023-05-25
Add `SymbolInformation.display_name`

olafurpg:

Previously, there was no reliable way to get the name of a symbol. You could parse it from the symbol descriptor but that only worked for global symbols (not locals) and was error-prone when the indexer chooses to a special encoding in the symbol for some reason (for example, all-lowercase because symbols are case-insensitive).

Now, clients can use `SymbolInformation.display_name` to reliably get the name of a symbol to render in API docs or some other application.

### Test plan

Green CI.


olafurpg:
cc/ @donsbot I'd love your input here.

varungandhi-src (approved):

olafurpg:
@donsbot good point. I updated the docstring to add an example to make it clear that the display name should not contain the fully-qualified scope.

PR #159, opened 2023-05-25
Add `SymbolInformation.signature_documentation`

olafurpg:

Previously, SCIP clients had to parse markdown from the `SymbolInformation.documentation` field to display only the signature information of a symbol (for example, method parameters and return type).

With the new `SymbolInformation.signature_documentation` field, indexers can now emit more structured information about the signature of a class/method including optional hyperlinks to the referenced symbols.

### Test plan

Green CI.


olafurpg:
cc/ @donsbot I'd love your thought on this approach.

I'm open for suggestions on an alternative encoding that `Document`. I think it's desirable to be able to support hyperlinking symbols in the signature since that's a pretty standard feature in API documentation tools like Javadoc.

scip.proto
```diff
@@ -309,6 +320,13 @@ message SymbolInformation {
   // - The symbol may encode names with special characters that should not be
   //   displayed to the user.
   string display_name = 6;
+  // (optional) The signature of this symbol as it's displayed in API
+  // documentation or in hover tooltips. For example, a Java method that adds
+  // two numbers this would have `Document.language = "java"` and `Document.text
+  // = "void add(int a, int b)". The `language` and `text` fields are required
+  // while other fields such as `Documentation.occurrences` can be optionally
+  // included to support hyperlinking referenced symbols in the signature.
+  repeated Document signature_documentation = 7;
```
varungandhi-src, line 329: Could you include an example of when it'd be useful for this field to have more than one element?

olafurpg: Good point, we can make it non-repeated. It's the same binary encoding either way so we can make it repeated in the future if there is demand for it.

varungandhi-src (approved):

scip.proto
olafurpg: Pushed a commit making it non-repeated

scip.proto
```diff
@@ -91,6 +91,15 @@ message Document {
...
   repeated SymbolInformation symbols = 3;
+
+  // (optional) Text contents of the this document. Indexers are not expected to
+  // include the text by default. It's preferrable that clients read the text
+  // contents from the file system by resolving the absolute path from joining
+  // `Index.metadata.project_root` and `Document.relative_path`. This field was
+  // introduced to support `SymbolInformation.signature_documentation`, but it
+  // can be used for other purposes as well, for example testing or when working
+  // with virtual/in-memory documents.
+  string text = 5;
```
tjdevries, line 102: what about "content" instead of "text"? don't have a strong preference either way.

olafurpg: I went with `text` to mirror `TextDocumentItem.text` in LSP (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem) and `Document.getText()` in VS Code. I'm personally fine either way but think it's a safe choice to stick with LSP terminology when appropriate.

PR #160, opened 2023-05-25
add new formatter

Numbers88s:
### Test plan
tested locally


efritz (approved):

bindings/go/scip/symbol_formatter.go
```diff
@@ -8,42 +8,71 @@ import (
...
+	IncludePackageManager:   func(_ string) bool { return false },
+	IncludePackageName:      func(_ string) bool { return false },
+	IncludePackageVersion:   func(_ string) bool { return false },
+	IncludeDescriptorString: func(_ string) bool { return true },
+	IncludeDescriptor:       func(_ *Descriptor) bool { return true },
+	IncludeDisambiguator:    func(_ string) bool { return true },
+}
+
+// SyntacticDescriptorOnlyFormatter formats only the syntactic descriptor part of the symbol.
+var SyntacticDescriptorOnlyFormatter = SymbolFormatter{
```
olafurpg, line 58: Can you elaborate on what "syntactic descriptor" means?

bindings/go/scip/symbol_formatter.go
```diff
@@ -8,42 +8,71 @@ import (
...
-	IncludePackageName    func(name string) bool
-	IncludePackageVersion func(version string) bool
-	IncludeDescriptor     func(descriptor string) bool
+	OnError                 func(err error) error
+	IncludeScheme           func(scheme string) bool
+	IncludePackageManager   func(manager string) bool
+	IncludePackageName      func(name string) bool
+	IncludePackageVersion   func(version string) bool
+	IncludeDescriptorString func(descriptor string) bool
+	IncludeDescriptor       func(descriptor *Descriptor) bool
```
olafurpg, line 17: This seems to be a breaking change. In general, I think it's desirable to avoid breaking changes since it's tiresome to consume bindings that regularly change their API. In this case, we could keep the old filter and add a new `IncludeParsedDescriptor(...)` filter

olafurpg: Not a blocking comment, just an observation. I don't think these bindings have high usage outside of Sourcegraph. Still, my general rule of thumb for public APIs is to never break them unless you absolutely have to.

efritz: `IncludeDescriptorString` is the old `IncludeDescriptor`. Renamed it, see new diff.

bindings/go/scip/symbol_formatter.go
```diff
@@ -44,6 +50,44 @@ var DescriptorOnlyFormatter = SymbolFormatter{
...
+	IncludeScheme:         func(scheme string) bool { return scheme == "local" },
+	IncludePackageManager: func(_ string) bool { return false },
+	IncludePackageName:    func(_ string) bool { return false },
+	IncludePackageVersion: func(_ string) bool { return false },
+	IncludeDescriptor:     func(_ string) bool { return true },
+	IncludeRawDescriptor:  includeReducedRawDescriptor,
+	IncludeDisambiguator:  func(_ string) bool { return false },
+}
+
+var reducedSuffixes = []Descriptor_Suffix{
```
olafurpg, line 76: Is it intentional to exclude parameters, type parameters, meta and macros? IIRC, the original goal was to only skip the method descriptor so that overloaded symbols would get the same symbol 🤔 The current "reduced" format seems a bit too ad-hoc to include in the official bindings.

olafurpg: Meant method *disambiguator*

olafurpg: Parameters and type parameters are very common suffixes. Is it intentional that we want the parameter of a method to be equal to the method itself?

PR #164, opened 2023-05-30
Add `SymbolInformation.enclosing_symbol`

olafurpg:

Previously, it was not possible to determine the enclosing symbol of a local because the syntax for local symbols has no hierarchy (unlike global symbols). This PR closes the gap by adding a field `SymbolInformation.enclosing_symbol` to allow documenting the "parent" or "owner" of a local symbol.

### Test plan

Green CI.


scip.proto
```diff
@@ -327,6 +327,17 @@ message SymbolInformation {
...
+  // symbols, the enclosing symbol should be parsed from the `symbol` field
+  // using the `Descriptor` grammar.
+  //
+  // The primary use-case for this field is to allow local symbol to be displayed
+  // in a symbol hierarchy for API documentation. It's OK to leave this field
+  // empty for local variables since local variables usually don't belong in API
+  // documentation. However, in the situation that you wish to include a local
+  // symbol in the hierarchy, then you can use `enclosing_symbol` to locate the
+  // "parent" or "owner" of this local symbol.
+  string enclosing_symbol = 8;
```
olafurpg, line 340: I debated using `owner` or `parent`, which is commonly the name of this functionality in compiler APIs that I have worked with (Java, Scala, C#, and TypeScript). However, I went with `enclosing_symbol` to make it more self-descriptive and also to reflect that I don't expect this to be a high-traffic API (unlike `owner` in the compiler that you use all the time).

olafurpg:
cc/ @donsbot this PR implements the last remaining proposal from https://github.com/sourcegraph/scip/issues/154

scip.proto
```diff
@@ -327,6 +327,17 @@ message SymbolInformation {
...
+  // (optional) The enclosing symbol if this is a local symbol.  For non-local
+  // symbols, the enclosing symbol should be parsed from the `symbol` field
+  // using the `Descriptor` grammar.
+  //
+  // The primary use-case for this field is to allow local symbol to be displayed
+  // in a symbol hierarchy for API documentation. It's OK to leave this field
+  // empty for local variables since local variables usually don't belong in API
+  // documentation. However, in the situation that you wish to include a local
+  // symbol in the hierarchy, then you can use `enclosing_symbol` to locate the
+  // "parent" or "owner" of this local symbol.
```
varungandhi-src, line 339: Reading this as someone writing an indexer, it's not clear under what circumstances I would want to include a local symbol in a symbol hierarchy. Could you add an example here?

varungandhi-src (approved):

varungandhi-src:
The Haskell bindings also need to be updated.

scip.proto
olafurpg: Good point. I added an example.

olafurpg:
The Haskell bindings are now automatically updated \o/

PR #171, opened 2023-06-13
go bindings: Don't panic on half-finished descriptors

efritz:
### Test plan

It doesn't crash now :clueless:

### DIFF
--- a/bindings/go/scip/symbol.go
+++ b/bindings/go/scip/symbol.go
@@ -96,7 +96,10 @@ func (s *symbolParser) error(message string) error {
 }
 
 func (s *symbolParser) current() rune {
-	return s.Symbol[s.index]
+	if s.index < len(s.Symbol) {
+		return s.Symbol[s.index]
+	}
+	return '\x00'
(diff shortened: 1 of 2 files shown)


Numbers88s (approved):

bindings/go/scip/symbol.go
```diff
@@ -168,7 +172,12 @@ func (s *symbolParser) parseDescriptor() (*Descriptor, error) {
 		default:
 		}
 	}
-	return nil, nil
+
+	end := s.index
+	if s.index > len(s.Symbol) {
+		end = len(s.Symbol)
+	}
+	return nil, errors.Newf("unrecognized descriptor %q", string(s.Symbol[start:end]))
```
efritz, line 180: Was there a reason this didn't return an error before?

tjdevries: probably just overlooked, don't think it was intentional

tjdevries (approved):
seems good to me. think we just missed the error there

PR #172, opened 2023-06-20
bindings/go: Add document-granularity streaming Index parser

varungandhi-src:
(I will update the generated files as a last step before merging.)

### Test plan

Added fuzz tests


bindings/go/scip/parse.go
```diff
@@ -0,0 +1,143 @@
...
+	ProcessDocument       func(*Document)
+	ProcessExternalSymbol func(*SymbolInformation)
+}
+
+// readVarint attempts to read a varint, using scratchBuf for temporary storage
+//
+// scratchBuf should be able to accommodate any varint size
+// based on its capacity, and be cleared before readVarint is called
+func readVarint(r io.Reader, scratchBuf []byte) (uint64, error) {
+	//buf := make([]byte, 10, 0)
```
varungandhi-src, line 27: TODO: Delete this

bindings/go/scip/parse.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+// IndexVisitor is a struct of functions rather than an interface since Go
+// doesn't support adding new functions to interfaces with default
+// implementations, so adding new functions here for new fields in
+// the SCIP schema would break clients. Individual functions may be nil.
+type IndexVisitor struct {
+	VisitMetadata       func(*Metadata)
+	VisitDocument       func(*Document)
+	VisitExternalSymbol func(*SymbolInformation)
```
varungandhi-src, line 19: I guess part of the question is, is this API fine for usage in the backend? I think it should be, as the backend can intern things when processing each Document, without having to perform any kind of ReadAll operation. Exposing an iterator style API with NextDocument() etc methods etc. is also possible, but that would be more complicated to use in Sourcegraph, with unclear benefits...

bindings/go/scip/parse.go
```diff
@@ -0,0 +1,149 @@
...
+			if dataLen > 0 {
+				numRead, err := r.Read(dataBuf)
+				if err != nil {
+					return errors.Wrapf(err, "failed to read data for %s", indexFieldName(fieldNumber))
+				}
+				if uint64(numRead) != dataLen {
+					return errors.Newf("expected to read %d bytes based on LEN but read %d bytes", dataLen, numRead)
+				}
+			}
+			if fieldNumber == metadataFieldNumber && pi.VisitMetadata != nil {
```
varungandhi-src, line 79: We could generalize some of this using reflection, but I don't think it's that complicated, and the Index structure is probably not going to be changing often enough to warrant the extra complexity of using reflection.

scip.proto
```diff
@@ -31,6 +31,7 @@ message Index {
   // external package won't get indexed for some reason then you can use this
   // field to provide hover documentation for those external symbols.
   repeated SymbolInformation external_symbols = 3;
+  // IMPORTANT: Update ParseStreaming on adding a new field here
```
olafurpg, line 34: Might be worth elaborating here.
```suggestion
  // IMPORTANT: when adding a new `Index` field, make sure to also update `ParseStreaming` in the Go bindings. See `IndexVisitor.VisitDocument()` as an example.
```

olafurpg (approved):
This is a great idea. I think the visitor pattern is ideal for use-cases like this.

scip.proto
varungandhi-src: Added more details.

bindings/go/scip/parse.go
efritz: If we can enforce an order here, metadata, then external symbols, then documents would be the best. The way we process today we build a map of external symbols that is available for each document.

efritz: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@a2c0e6a8bf5e55227c0bca7db2e4ac0a20c1162d/-/blob/enterprise/internal/codeintel/uploads/internal/background/processor/scip.go?L30:6

varungandhi-src: > If we can enforce an order here, metadata, then external symbols, then documents would be the best

Protobuf doesn't actually guarantee that. The tag length value representation means that fields can be interleaved.

https://protobuf.dev/programming-guides/encoding/

So the parser can't really ensure that without actually buffering the input. On the consumer side, there are a couple of options:

1. Buffer compressed versions of the documents (compressed because string interning will likely reduce memory usage), and process the external symbols as they arrive.
2. Do two traversals over the index, first for the external symbols only, and then for the documents only (or vice-versa, if needed).

efritz: Ah ok two traversals wouldn't be bad, just set up two visitors one with a no-op for documents and one with a no-op for symbols.

Makes sense!

efritz (approved):

PR #176, opened 2023-06-22
bindings/go: Fix bug with incomplete reads

varungandhi-src:
### Test plan

Added regression test


tjdevries:
Did this go test failure just start happening? or was this from before?

varungandhi-src:
That failure is entirely unrelated to this PR AFAICT, not sure why it's coming up but will look into it tomorrow before merging.

varungandhi-src:
Perhaps an artifact of targeting a branch other than `main`, doesn't seem to be any issue any more.

PR #181, opened 2023-06-28
bindings/go: Add memory usage test for streaming parser

varungandhi-src:
We rely on low memory usage in Sourcegraph. https://github.com/sourcegraph/sourcegraph/pull/53828

So add a test to avoid regressing memory usage.

### Test plan

Added a new test

### DIFF
--- a/bindings/go/scip/memtest/low_mem_test.go
+++ b/bindings/go/scip/memtest/low_mem_test.go
@@ -20,6 +20,7 @@ func TestLowMemoryParsing(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(tmpFile.Name())
 
+	// Total index size will be about (textSize + ε) * docCount ~= 128 MB
 	const textSize = 128 * 1024
 	const docCount = 1000
 	{
--- a/bindings/go/scip/memtest/low_mem_test.go
+++ b/bindings/go/scip/memtest/low_mem_test.go
@@ -15,6 +15,9 @@ import (
 	"github.com/sourcegraph/scip/bindings/go/scip"
 )
 
+// Do not add any other tests in this sub-package, so that the
+// SetMemoryLimit call doesn't interfere with other running tests
+
 func TestLowMemoryParsing(t *testing.T) {
 	tmpFile, err := os.CreateTemp(t.TempDir(), "very-large-index.scip")
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


bindings/go/scip/memtest/low_mem_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/sourcegraph/scip/bindings/go/scip"
+)
+
+func TestLowMemoryParsing(t *testing.T) {
+	tmpFile, err := os.CreateTemp(t.TempDir(), "very-large-index.scip")
+	require.NoError(t, err)
+	defer os.RemoveAll(tmpFile.Name())
+
+	const textSize = 128 * 1024
+	const docCount = 1000
```
efritz, line 24: Can we get a rough estimation in a comment of the resulting index size from these values? Tried to gauge it at-a-glance but the strings/repeat having an indeterminate number of characters fuzzed me up. Lower bound is good as well.

bindings/go/scip/memtest/low_mem_test.go
```diff
@@ -0,0 +1,59 @@
...
+		_, err = tmpFile.Write(indexBytes)
+		require.NoError(t, err)
+
+		_, err = tmpFile.Seek(0, io.SeekStart)
+		require.NoError(t, err)
+
+		runtime.GC()
+	}
+
+	require.Greater(t, docCount, 100)
```
efritz, line 43: Oh man this is so smart+defensive, thanks for this.

bindings/go/scip/memtest/low_mem_test.go
```diff
@@ -0,0 +1,59 @@
...
+	}}
+
+	// No OOM
+	err = indexVisitor.ParseStreaming(tmpFile)
+	_ = curDoc
+	require.NoError(t, err)
+}
+
+// Do not add any other tests in this sub-package, so that the
+// SetMemoryLimit call doesn't interfere with other running tests
```
efritz, line 59: Might be worth adding this to the top of the file or in a very obvious sibling file in this package.

efritz (approved):
Comments are just doc nits

bindings/go/scip/memtest/low_mem_test.go
varungandhi-src: Index size should be about `(textSize + δ) * docCount` where δ < 100 probably based on carrying some integers related to offsets and lengths. So this will end up being about 128 MB.

varungandhi-src: Added a comment.

efritz (approved):

PR #186, opened 2023-07-12
formatter: Fix escaping logic

efritz:
We were failing to escape the SCIP identifier names on formatting.

### Test plan

New unit tests


olafurpg (approved):
LGTM 👍🏻 This is a great fix! I've noticed that we don't correctly escape in the snapshot output

PR #188, opened 2023-07-19
Stop sorting CLI options

abitrolly:
Fixed in https://github.com/urfave/cli/releases/tag/v2.19.2

### Test plan
```sh
scip snapshot -h
```

**Before**
```
OPTIONS:
   --comment-syntax value  Comment syntax to use for snapshot files (default: "//")
   --from value            Path to SCIP index file (default: "index.scip")
   --help, -h              show help (default: false)
   --project-root value    Override project root in the SCIP file. For example, this can be helpful when the SCIP index was created inside a Docker image or created on another computer
   --strict                If true, fail fast on errors (default: true)
   --to value              Path to output directory for snapshot files (default: "scip-snapshot")
```

**After**
```
OPTIONS:
   --from value            Path to SCIP index file (default: "index.scip")
   --to value              Path to output directory for snapshot files (default: "scip-snapshot")
   --project-root value    Override project root in the SCIP file. For example, this can be helpful when the SCIP index was created inside a Docker image or created on another computer
   --strict                If true, fail fast on errors (default: true)
   --comment-syntax value  Comment syntax to use for snapshot files (default: "//")
   --help, -h              show help (default: false) 
```

### DIFF
--- a/cmd/main_test.go
+++ b/cmd/main_test.go
@@ -45,7 +45,7 @@ func TestCLIReferenceInSync(t *testing.T) {
 		require.Nil(t, err)
 		help := strings.TrimSpace(string(helpBytes))
 		require.Truef(t, strings.Contains(readme, help),
-			"Readme.md missing help text %s for %s.\nRun `%s` and paste the output in the Readme.",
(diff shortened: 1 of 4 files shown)


olafurpg (approved):
LGTM 👍🏻 Thank you!

PR #189, opened 2023-07-21
Fix protobuf generation

olafurpg:
Previously, the build was failing because of an HTTP error connecting to the buf plugin registry. This commit fixes the problem by replacing the buf Go plugin with the standard Go protobuf generator.

### Test plan
Green CI.


dev/proto-generate.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -xeuo pipefail
```
olafurpg, line 3: Added `-x` to make it easier to debug what command failed

varungandhi-src (approved):

PR #190, opened 2023-07-24
[WIP] fix: Flakiness in TypeScript CI

varungandhi-src:
### Test plan

Run CI

### DIFF
--- a/.github/workflows/typescript.yml
+++ b/.github/workflows/typescript.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - '.github/workflows/**'
+      - 'bindings/typescript/**'
       - '**.ts'
       - '**/package.json'
       - '**/tsconfig.json'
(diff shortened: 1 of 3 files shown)


varungandhi-src:
OK, the CI is not running properly, because I'm seeing a failure locally...

varungandhi-src:
This is the failure I'm getting locally.

```
❯ ./node_modules/.bin/tsc -b bindings/typescript
bindings/typescript/scip.ts:1753:20 - error TS2352: Conversion of type 'import("/Users/varun/Code/scip/bindings/typescript/scip").scip.DiagnosticTag' to type 'DiagnosticTag[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

1753             return pb_1.Message.getFieldWithDefault(this, 5, DiagnosticTag.UnspecifiedDiagnosticTag) as DiagnosticTag[];
```

The TypeScript version is correct in CI. Some other possibilities:
- Some other configuration on my local machine is getting picked up, causing this failure locally but not in CI.
- There is some other configuration in CI causing the type error to not trigger a failure...

varungandhi-src:
Closed in favor of https://github.com/sourcegraph/scip/pull/193

Issue #149, opened 2023-05-02
CI: Typescript CI is flaky


keynmol:
Failed job: https://github.com/sourcegraph/scip/actions/runs/4860856491/attempts/1
Re-run fixed it immediately.

Commit itself has nothing to do with code.

PR #192, opened 2023-07-24
Disable typechecking of autogenerated scip.ts bindings

olafurpg:
Attempt to fix https://github.com/sourcegraph/scip/issues/149

### Test plan
Green CI.


olafurpg:
Closing this since there are no remaining files to typecheck if we ignore scip.ts. The actual fix would be to remove the typescript typechecking job, if we want to go this route.

PR #195, opened 2023-08-03
add solidity definitions

zfy0701:


scip.proto
```diff
@@ -309,7 +309,12 @@ message SymbolInformation {
       Union = 59;
       Value = 60;
       Variable = 61;
-      // Next = 62;
+      // For Solidity
+      Contract = 62;
+      Library = 63;
+      Modifier = 64;
+      Error = 65;
```
varungandhi-src, line 316: Could you please move these earlier to be in alphabetical order? Similar to the ordering in the languages enum, we'd like this to be in alphabetical order rather than numeric order, that way it's easier to check if a particular construct is present.

zfy0701: done

zfy0701: done, please take another look

varungandhi-src:
Merged. Thanks for your contribution!

PR #208, opened 2023-10-24
Add MethodSpec kind

jtibshirani:
This change adds a new kind representing a method specification on an
interface.

The motivation: in some cases, it can be helpful to distinguish method specs
from method implementations. For example, when searching for symbols, method
specs should usually be ranked higher than method implementations.

### Test plan

Green CI!


scip.proto
```diff
@@ -265,6 +265,8 @@ message SymbolInformation {
       // Analogous to 'ThisParameter' and 'SelfParameter', but for languages
       // like Go where the receiver doesn't have a conventional name.
       MethodReceiver = 27;
+      // A method specification on an interface
+      MethodSpec = 66;
```
olafurpg, line 269: Should we update the comment to mention Go? This change seems specifically motivated for Go and I'm not aware of another language that uses the terminology "method specification". It's more common to talk about "abstract method" or "interface method". I'll leave it to @varungandhi-src or @keynmol to stamp.

varungandhi-src: I agree, let's mention `For Go` or similar like the other comments.

Couple of other minor suggestions (not strictly required):

1. Other cases in this type don't use shorthands and the official Go docs use the full spelling "method specification", so I'd weakly prefer following that and spelling it out in full as "MethodSpecification" rather than "MethodSpec".

2. It'd be nice to also add all the similar constructs in other languages at the same time.

    ```
    // For Java, Kotlin
    AbstractMethod
    // For Rust, Scala
    TraitMethod
    // For C++
    PureVirtualMethod
    // For Swift, Objective-C
    ProtocolMethod
    // For Haskell, Purescript etc.
    TypeclassMethod
    ```
    Since the enum constants here are meant to match in-language terminology, it makes sense to add separate constants for different languages.

varungandhi-src (approved):
Approving since the existing comments are non-blocking/can be fixed later.

scip.proto
varungandhi-src: I've pushed the suggested changes to this branch.

jtibshirani:
Thanks @varungandhi-src for fixing and merging! I have a better understanding of the process now.

PR #209, opened 2023-10-25
schema: Add more symbol kinds for Ruby, C# and C++

varungandhi-src:
Based of points noted in https://github.com/sourcegraph/sourcegraph/pull/57879#issuecomment-1778239962 as well as just noting down related types

### Test plan

n/a


olafurpg (approved):
Thank you!

PR #214, opened 2023-10-31
feat: Add flag for turning off colors in TTY printing

varungandhi-src:
It is a bit annoying to copy the colored output into something like Markdown.

### Test plan

Manually checked.

With `scip print` or `scip print --color=true`:

image

With `scip print --color=false`

image


keynmol:
Given you're already touching this code, would be good to follow the CLI conventions (https://clig.dev/#output) and at least handle NO_COLOR env variable as well: 

```
Use color with intention. For example, you might want to highlight some text so the user notices it, or use red to indicate an error. Don’t overuse it—if everything is a different color, then the color means nothing and only makes it harder to read.

Disable color if your program is not in a terminal or the user requested it. These things should disable colors:

    stdout or stderr is not an interactive terminal (a TTY). It’s best to individually check—if you’re piping stdout to another program, it’s still useful to get colors on stderr.
    The NO_COLOR environment variable is set.
    The TERM environment variable has the value dumb.
    The user passes the option --no-color.
    You may also want to add a MYAPP_NO_COLOR environment variable in case users want to disable color specifically for your program.
```

varungandhi-src:
Added handling for `NO_COLOR`.

cmd/scip/print.go
```diff
@@ -12,31 +13,49 @@ import (
...
 			if indexPath == "" {
 				return errors.New("missing argument for path to SCIP index")
 			}
-			return printMain(indexPath, json, c.App.Writer)
+			// Following https://no-color.org/
+			if val, found := os.LookupEnv("NO_COLOR"); found && val != "" {
+				switch val {
+				case "":
+					break
+				case "0", "false", "FALSE", "off", "OFF":
```
keynmol, line 46: Perhaps it's easier to just lowercase the value before matching

varungandhi-src: Ah yes, done

keynmol (approved):

Issue #131, opened 2023-03-16
Add a (forward) declaration role


varungandhi-src:
C++ code can have forward declarations; it'd be useful to track that separately, so that we can be smarter in returning results. For example, in Find References, we can potentially separate out forward declarations, or reduce the ranking score.

I think we should add a new SymbolRole for forward declarations.

PR #217, opened 2023-11-03
feat: Add SymbolRole for representing forward declarations

varungandhi-src:
Adds a new `SymbolRole` for forward declarations so that certain occurrences can be marked
with this role for C-based languages, optionally allowing clients such as a web UI to filter out
or down-rank them.

Question:
- [x] There are similar constructs in other languages, most prominently type signatures in OCaml's `.mli` files. Should we change the name `ForwardDeclaration` to be something more generic else to account for that too? Should we add a separate `SymbolRole` for that? Deal with it later instead of right now? -- Renamed to `ForwardDefinition`.

Fixes https://github.com/sourcegraph/scip/issues/131

Related PR: https://github.com/sourcegraph/sourcegraph/pull/58082

### Test plan

I've added an optional `forward_decl` field to reprolang along with a snapshot test for it.


mrnugget (approved):

scip.proto
```diff
@@ -474,6 +474,11 @@ enum SymbolRole {
   Generated = 0x10;
   // Is the symbol in test code?
   Test = 0x20;
+  // Is this a forward declaration for the symbol?
+  //
+  // Applies to languages like C, C++ and Objective-C
+  // Forward declarations should not use the Definition role.
+  ForwardDeclaration = 0x40;
```
olafurpg, line 481: Given we have a `Definition` role, what do you think about calling this a `ForwardDefinition`? This would avoid the need to add separate kinds of roles for different languages like OCaml.

varungandhi-src: Done.

cmd/scip/tests/snapshots/output/forward-def/forward_def.repro
```diff
@@ -0,0 +1,8 @@
+ reference forward_definition abc#
```
olafurpg, line 1: nit: I would expect the repro syntax to work like this
```suggestion
forward_definition abc#
```

olafurpg (approved):
LGTM 👍🏻 One non-blocking comment

mrnugget:
@olafurpg Can I merge? @varungandhi-src said he's going to be out rest of day and asked wehther I could merge this.

olafurpg:
No problem merging, the comment was only minor

cmd/scip/tests/snapshots/output/forward-def/forward_def.repro
varungandhi-src: I put it inside `reference` so that it's clearer that it is also a reference, not a tri-state of definition/reference/forward definition.

olafurpg: Should we update the docstring to emphasize that you should set `Definition` and `ForwardDefinition`? It might be easy to misunderstand that setting `ForwardDefinition` alone implies `Definition`

varungandhi-src: No, that would be incorrect. You should NOT set `Definition` if you set `ForwardDefinition` because a `ForwardDefinition` is actually a reference, it is not a Definition because it doesn't have a body.

I've added a lint for this in the `scip lint` command and mentioned the above requirement in the doc comment.

olafurpg: It is a definition IMO because it's a valid destination for "Go to definition". When I trigger "Go to implementation" then I would expect it to go to the non-forward definition.

varungandhi-src: My 2c based on working in C++: for Go to Definition, I always want to be taken to the method body if available, not to the forward declaration. Yes, if the Definition isn't available, then a forward declaration is OK as a fallback.

The "implementation" terminology is a bit confusing here, because with Find Impls, there are different possible Impls for the same interface & there is some kind of subtyping involved. However, with forward declarations, there should only be one place where the body of the method exists, multiple can't exist, so lumping the body of the method along with corresponding methods in subclasses (in C++, doesn't apply to C) would be confusing.

varungandhi-src: If you still disagree, let's discuss it synchronously some time next week. It would be helpful for me to show some live examples of code nav behavior in C++.

olafurpg: Or updating the docstring to emphasize that `Definition` should not be set

varungandhi-src: Ugh, I thought I updated the docstring earlier, but I must've stashed that change instead of merging it. 🤦🏽 

Yes, I'll do that.

olafurpg: In my mental model, it is a definition with lower priority. If you have two ambiguous definitions and one has the `ForwardDefinition` role then we can discard it. I would rather use that application logic than requiring the client to treat a non-definition (which is confusingly named `ForwardDefinition`) as a definition

PR #224, opened 2023-12-12
schema: Specify the encoding for character offsets

varungandhi-src:
This PR adds a new `PositionEncoding` field so that indexers can specify
what type of character offsets they are using. This way, consumers of SCIP
can unambiguously interpret the offsets for non-ASCII data.

I have kept it as a field on `Document` rather than `Index` because:

1. There is no additional benefit from having it on `Index` because
    occurrences only belong inside Documents, not outside.
2. It allows one to concatenate indexes from different sources
    which use different kinds of offsets.

### Test plan

n/a


scip.proto
```diff
@@ -102,8 +103,37 @@ message Document {
   // can be used for other purposes as well, for example testing or when working
   // with virtual/in-memory documents.
   string text = 5;
+
+  // Specifies the encoding used for source ranges in this Document.
+  PositionEncoding position_encoding = 6;
```
olafurpg, line 108: I'm open to adding a comment with recommendations. Something like "If you are unsure which encoding to use, we recommend `UTF16CodeUnitOffsetFromLineStart` if you are working with JS/JVM and `UTF8CodeUnitOffsetFromLineStart` if you working with Python". Or at least something about how to approach deciding which one to use. Not a blocking comment.

olafurpg (approved):

olafurpg:
Good point about making this per document. Would make sense to support the same for the text document encoding.

scip.proto
varungandhi-src: Added a recommendation.

PR #220, opened 2023-11-05
Included diagnostics in `scip snapshot` output

matthewnitschke-wk:
This concept originally came out of [this issue](https://github.com/Workiva/scip-dart/issues/98), which was individually reported in `sourcegraph/scip` within [this issue](https://github.com/sourcegraph/scip/issues/213)

Adds the option to include indexed diagnostics to the results of `scip snapshot`

This functionality is explicitly enabled via the `--include-diagnostics` flag, to support backwards compatibility with current  snapshot tests (eg, the inclusion of `diagnostics` in snapshot files is opt-in)

Example format is as follows:

- Given this example dart file
    ```dart
    void _unusedElement() {}
    ```
- Running `scip snapshot --include-diagnostics` will result in
   ```dart
      void _unusedElement() {}
   //      ^^^^^^^^^^^^^^ definition scip-dart pub test_package lib/main.dart _unusedElement().
  //      diagnostic UNUSED_ELEMENT
   ```

### Test plan
- I've verified the functionality myself on various indexed scip files, but when attempting to write some unit tests around the new functionality, I was a bit confused on the current test setup, specifically around `repolang` and how to add diagnostic concepts to it. I'm open to suggestions / contributions to the pr addressing this missing test coverage

PR #226, opened 2023-12-18
Include diagnostics in snapshot output

### LINKED ISSUE #213: Include diagnostics in `scip snapshot` output
See: https://github.com/Workiva/scip-dart/issues/98

It would be nice to include this in the output, there shouldn't be a need to implement this on a per indexer basis.

jtibshirani:
This change adds diagnostic information to the snapshot output to support
testing. To make it possible to test this in the SCIP repo, it updates the
reprolang generator to output a diagnostic for every identifier that starts
with 'deprecated'.

Example output:
```
 reference deprecatedMethod.
#          ^^^^^^^^^^^^^^^^^ reference diagnostics.repro/deprecatedMethod.
#          diagnostic Warning: deprecated identifier
```

Closes #213

### Test plan

New snapshot test to check diagnostics output

### DIFF
--- a/bindings/go/scip/testutil/format.go
+++ b/bindings/go/scip/testutil/format.go
@@ -17,7 +17,6 @@ import (
 // that is suitable for snapshot testing.
 func FormatSnapshots(
 	index *scip.Index,
-	includeDiagnostics bool,
 	commentSyntax string,
 	symbolFormatter scip.SymbolFormatter,
 	customProjectRoot string,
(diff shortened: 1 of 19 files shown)


jtibshirani:
This PR is based on https://github.com/sourcegraph/scip/pull/220, with the following differences:
* Don't introduce a new flag `--include-diagnostics`, as I think it's fine to just enable this by default. When people update to this new SCIP version, they can just update their test snapshots if needed.
* Print the diagnostic warning and message instead of the 'code', as these are required fields and more likely to be present

varungandhi-src:
Nice work. I noticed that we were truncating the multiline output for docs (and you probably copied that for the diagnostic message), which seems like undesirable behavior from the POV of an indexer as you want to make sure the full contents are correct. So I've tweaked that.

varungandhi-src:
Rebased on include the build failure fix from https://github.com/sourcegraph/scip/pull/228

PR #231, opened 2024-02-06
Add support for `SymbolInformation.signature`

olafurpg:


olafurpg:
* On a second thought, having separate `Constant` message allows code reuse between types and terms. It's especially desirable to add terms for annotation arguments.
* We should consider adding `TypeSignature.value` or something equivalent for basic `type A = B` signatures. The current lower/upper bound design is to support `type A >: B` in Scala, but we should not sacrifice ergonomics for the >99% use-case.

olafurpg:
Closing since this isn't going to land in the near future.

PR #236, opened 2024-04-21
Test Command

### LINKED ISSUE #235: `scip test` command
While developing tree-sitter grammars, I've gotten pretty used to the `tree-sitter test` command, where selections are explicitly stated within test files, and the result passes and fails based on whether the parser matches whats specified

What would the thought be around creating a `scip test` command which mirrors how tree-sitter highlight tests work?

For example, the following file would be committed a `test/` directory, and `scip test ./index.scip` would pass or fail depending on if the `Foo` symbol had the correct definition
```dart
class Foo {
//    ^ definition scip-dart pub some_package 1.0.0 lib/`main.dart`/main().
}
```

Ideally all "selectors" could be used this way as well, `reference`, `documentation`, `diagnostic`, and `relationship`...

This would be offered as an alternate to the existing `scip snapshot`, where individual language features are tested, making it clearer from a test file perspective what is intentional

matthewnitschke-wk:
Implements the proposal specified here: #235 

Closes #235


docs/CLI.md
```diff
@@ -112,6 +112,58 @@ OPTIONS:
...
+   file matches manually specified symbol fields within an index. This can be used in
+   conjunction with snapshot to provide a cohesive testing framework for scip indexers.
+
+OPTIONS:
+   --from value            Path to SCIP index file (default: "index.scip")
+   --comment-syntax value  Comment syntax to use for snapshot files (default: "//")
+   --help, -h              show help
+```
+
+### File Specification
```
matthewnitschke-wk, line 135: Looks like this file `docs/CLI.md` is used in a unit test to validate the output of `scip --help`

I'm open to suggestions on a better location for this documentation

docs/CLI.md
```diff
@@ -112,6 +112,65 @@ OPTIONS:
...
+Given the following example, the cli command `scip test` will exit with a non-zero exit code if the provided `index.scip` file does not contain a definition, at the specified range, with the specified symbol
+```js
+function someFunction() {
+   //    ^^^^^^^^^^^^ definition scip-typescript npm test_package 1.0.0 lib/`test.js`/someFunction().
+}
+```
+
+Three range selection comment formats are supported:
+- `// ^^^`: enforces the length of the symbol. Will fail if the range at this location does not equal 3 characters
+- `// ^`: ignore length, symbol can occur at any point between the start and end of the `^` character
```
varungandhi-src, line 148:
```suggestion
- `// ^`: ignore length, `^` can occur at any point to any character in the occurrence
```

docs/CLI.md
```diff
@@ -112,6 +112,65 @@ OPTIONS:
...
+
+Given the following example, the cli command `scip test` will exit with a non-zero exit code if the provided `index.scip` file does not contain a definition, at the specified range, with the specified symbol
+```js
+function someFunction() {
+   //    ^^^^^^^^^^^^ definition scip-typescript npm test_package 1.0.0 lib/`test.js`/someFunction().
+}
+```
+
+Three range selection comment formats are supported:
+- `// ^^^`: enforces the length of the symbol. Will fail if the range at this location does not equal 3 characters
```
varungandhi-src, line 147:
```suggestion
- `// ^^^`: enforces the length of the occurrence. Will fail if the range at this location does not equal 3 characters
```

docs/CLI.md
```diff
@@ -112,6 +112,65 @@ OPTIONS:
...
+- `// ^`: ignore length, symbol can occur at any point between the start and end of the `^` character
+- `// <-`: ignore length, and treat the "start" of the symbol at the beginning of the comment. This follows sublime-text's format
```
varungandhi-src, line 149:
```suggestion
- `// <-`: ignore length, and treat the character above the first comment character as the start of the occurrence, similar to Sublime Text
```

docs/CLI.md
```diff
@@ -112,6 +112,65 @@ OPTIONS:
...
+
+OPTIONS:
+   --from value            Path to SCIP index file (default: "index.scip")
+   --comment-syntax value  Comment syntax to use for snapshot files (default: "//")
+   --help, -h              show help
+```
+
+### File Specification
+
+The validation files for `scip test` are loosely based on [sublime text's](https://www.sublimetext.com/docs/syntax.html#testing) syntax highlighting files. Symbol ranges are selected using `^`, `^^` and `// <-`. Symbol attributes can be specified as the first word: `definition`, `reference`, `diagnostic`, and all subsequent lines are interpreted based on the kind of symbol attribute you are writing a test for
```
varungandhi-src, line 137:
```suggestion
The validation files for `scip test` are loosely based on [Sublime Text's](https://www.sublimetext.com/docs/syntax.html#testing) syntax highlighting files. Symbol ranges are selected using `^`, `^^` and `// <-`. Symbol attributes can be specified as the first word: `definition`, `reference`, `diagnostic`, and all subsequent lines are interpreted based on the kind of symbol attribute you are writing a test for
```

Here, `diagnostic` is technically a field and definition/reference are roles. 'symbol attributes' sounds like standard SCIP terminology, but it actually doesn't have any meaning in SCIP. Later in this doc, the same thing is referred to as 'selection kind'. Alternate terminology suggestion, say something like:

> There are 2 types of checks supported by `scip test` - for symbol roles and for diagnostics. ...
>
> The syntax to check symbol roles is as follows: \<code snippet\>
>
> The syntax to check diagnostics is as follows: \<code snippet\>

If you want, it might make sense to add a way to check/test relationships too.

cmd/scip/test.go
```diff
@@ -0,0 +1,369 @@
...
+// This can be a definition, reference, documentation, or diagnostic
+type symbolAttribute struct {
+	// the column number where this symbol starts
+	start int
+
+	// the length of the symbol's name
+	length int
+
+	// the type of attribute that this is
+	kind string
```
varungandhi-src, line 107: Please make this an enum, not a `string`

cmd/scip/test.go
```diff
@@ -0,0 +1,369 @@
...
+	enforceLength bool
+}
+
+// commentsForLine returns the list of lines, after a provided [lineNumber], which are
+// classified as comment. The comment type can be configured using [commentSyntax].
+//
+// Returns the list of symbolAttributeTestCase(s) for the provided line, and the number of
+// of lines that were "consumed" by the cases on this line
+func testCasesForLine(lineNumber int, lines []string, commentSyntax string) ([]*symbolAttributeTestCase, int) {
+	testCases := []*symbolAttributeTestCase{}
```
varungandhi-src, line 132: AFAICT there is no need to use pointers here since the `nil` case isn't possible so it'd be simpler to remove the extra `*` and `&`

cmd/scip/test.go
```diff
@@ -0,0 +1,369 @@
...
+		return testCases, 0
+	}
+
+	testLines := []*symbolAttributeTestCase{}
+	usedLines := 0
+	for i := lineNumber + 1; i < len(lines); i++ {
+		line := lines[i]
+
+		if !strings.HasPrefix(strings.TrimSpace(line), commentSyntax) {
+			// if the line does not start with a comment, we're done. break
```
varungandhi-src, line 146:
```suggestion
```

The line of code above is easy to understand.

cmd/scip/test.go
```diff
@@ -0,0 +1,369 @@
...
+func isValidTestCase(testCase *symbolAttributeTestCase, attributes []*symbolAttribute) bool {
+	for _, attr := range attributes {
+		if isValidTestCaseForAttribute(testCase, attr) {
+			return true
+		}
+	}
+	return false
+}
+
+func isValidTestCaseForAttribute(testCase *symbolAttributeTestCase, attr *symbolAttribute) bool {
```
varungandhi-src, line 273: These should probably be methods on `symbolAttributeTestCase` called `CheckAll` and `Check` respectively.

The prefix 'isValid' makes it sound like we're doing a well-formedness check, whereas we're actually doing a _comparison_ between the SCIP index and the test data.

cmd/scip/test_test.go
```diff
@@ -0,0 +1,208 @@
...
+			start:  3,
+			length: 5,
+			kind:   "definition",
+			data:   "scip-dart pub somePackage 1.0.0 lib/`foo.dart`/bar().",
+		},
+	)
+	require.False(t, notEqualSymbol)
+}
+
+func TestIsValidTestCaseForAttribute_additionalData(t *testing.T) {
```
varungandhi-src, line 151: Can we reuse the code we have for reprolang (to easily generate a SCIP index with the desired data) as well as the test parsing here (to generate the test cases) so that the bulk of tests here involve end-to-end testing and avoiding having to write a bunch of struct literals by hand?

varungandhi-src (commented):
Looks good in broad strokes, but I think there is a bunch of room to simplify things

matthewnitschke-wk:
@varungandhi-src amusingly long delay on this one, but finally got around to make your suggestions

Should be ready for another round of reviews!

docs/CLI.md
varungandhi-src: I realize I recommended linking to the `main` branch earlier, but it's probably better to link to the URL for the released version so that there is less risk of confusion in case an old version of the subcommand doesn't support some syntax supported by a new version/on main. I've pushed a change for that.

cmd/scip/main_test.go
```diff
@@ -101,3 +104,116 @@ func TestSCIPSnapshots(t *testing.T) {
...
+		dir        string
+		passOutput autogold.Value
+		failOutput autogold.Value
+	}
+
+	// To update the snapshot values, run 'go test ./cmd/scip -update'.
+	testCases := []TestCase{
+		{"roles",
+			autogold.Expect("✓ passes.repro (3 assertions)\n"),
+			autogold.Expect(`✗ fails-wrong-role.repro
```
varungandhi-src, line 136: @matthewnitschke-wk I've added snapshot tests here for checking the output, so that we don't accidentally regress something (earlier, this test was just checking for pass/fail instead of checking the output).

cmd/scip/test_test.go
```diff
@@ -0,0 +1,58 @@
+package main
+
+import (
+	"slices"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestTestCasesForLine(t *testing.T) {
```
varungandhi-src, line 11: @matthewnitschke-wk Thank you for adding this test, it compactly expresses the different kinds of syntax that are available and how they get parsed. Nice!

varungandhi-src (approved):

varungandhi-src:
@matthewnitschke-wk I've cut a new v0.5.0 release with the test subcommand.

PR #247, opened 2024-06-11
schema: Document that ranges should be half-open

varungandhi-src:
### Test plan

n/a, codegen checked by CI


kritzcreek:
While we're defining things. Can Ranges be empty? As in Range(p1, p2) where p1 == p2?

varungandhi-src:
Yes, ranges can be empty.

kritzcreek (approved):

PR #248, opened 2024-06-11
bindings/go: Add common APIs for Range and Position

varungandhi-src:
The Range type in this package is 128-bits, whereas some Range types
in the Sourcegraph packages are 256-bits. Additionally, some of the APIs
would previously work using `*Range` instead of just `Range`.

This PR fixes the pointer-based APIs and adds commonly used APIs
for Range and Position so we don't have to keep re-implementing them.

We also add documentation that ranges are half-open, not closed.

### Test plan

Added new tests

### DIFF
--- a/bindings/go/scip/position.go
+++ b/bindings/go/scip/position.go
@@ -45,11 +45,97 @@ func (p Position) Less(other Position) bool {
 	return p.Character < other.Character
 }
 
-// NewRange converts an SCIP range into `Range`
+func NewRange(scipRange []int32) (Range, error) {
+	switch len(scipRange) {
+	case 3:
+		startLine := scipRange[0]
+		startChar := scipRange[1]
+		endChar := scipRange[2]
+		if startLine >= 0 && startChar >= 0 {
+			if endChar >= startChar {
+				return Range{
+					Start: Position{
+						Line:      startLine,
+						Character: startChar,
+					},
+					End: Position{
+						Line:      startLine,
+						Character: endChar,
+					},
+				}, nil
+			}
+			return Range{}, EndBeforeStartRangeError
+		}
+		return Range{}, NegativeOffsetsRangeError
+	case 4:
+		startLine := scipRange[0]
+		startChar := scipRange[1]
+		endLine := scipRange[2]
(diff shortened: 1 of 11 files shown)


bindings/go/scip/position.go
```diff
@@ -16,18 +19,131 @@ type Position struct {
...
+	if p.Line < other.Line {
+		return true
+	}
+	if p.Line > other.Line {
+		return false
+	}
+	return p.Character < other.Character
+}
+
+func NewRange(scipRange []int32) (Range, error) {
```
varungandhi-src, line 48: TODO: Add some tests for this

bindings/go/scip/position_test.go
```diff
@@ -0,0 +1,156 @@
...
+	require.Truef(t, r1.Intersects(r2), "%+v.Intersects(%+v)", r1, r2)
+	require.Truef(t, r2.Intersects(r3), "%+v.Intersects(%+v)", r2, r3)
+	require.Falsef(t, r1.Intersects(r3), "%+v.Intersects(%+v)", r1, r3)
+}
+
+func genPosition() *rapid.Generator[Position] {
+	return rapid.Custom(func(t *rapid.T) Position {
+		return Position{
+			Line:      rapid.Int32Range(0, 10).Draw(t, "Line"),
+			Character: rapid.Int32Range(0, 10).Draw(t, "Character"),
```
kritzcreek, line 68: Oh, I made these ranges really small to trigger a case I wanted to see fail. I think these should be `rapid.Int32Range(0, math.MaxInt32)` (Assuming `rapid.Int32Range` is `[max..min]` I couldn't find documentation)

kritzcreek: Or maybe `(0..100)` if we're worried we otherwise rarely get ranges that interact with each other...

varungandhi-src: We're not doing any math, so I don't think it's worth testing the full range. Rather, more tests in a smaller range are reasonably likely to exercise all the code paths.

kritzcreek (approved):

PR #249, opened 2024-06-11
codegen: Pass --installdir flag to cabal install

varungandhi-src:
In some configurations, 'cabal install' will install the binary
into the current directory instead of in the global directory
on 'PATH', which can cause the haskell step to fail. Pass
the --installdir flag to avoid that.

### Test plan

Covered by existing codegen


kritzcreek (approved):
Don't have a Haskell toolchain around I can test this with, but from reading the docs this looks good.

PR #257, opened 2024-06-25
Dockerised binding regeneration

keynmol:
Fixes GRAPH-709

Several of us had problems with local environment trying to regenerate bindings.
We're switching to the dockerised build as a stopgap solution.

### Test plan

- new script will be used on CI


Dockerfile.bindings
```diff
@@ -0,0 +1,33 @@
+FROM ubuntu:jammy
+
+RUN apt-get update -q && apt-get install -y git curl xz-utils tar gpg build-essential
```
varungandhi-src, line 3: Should we pin the image to a SHA and avoid `apt-get update`? I'm wondering if we might run into differences because of that over time.

keynmol: I think apt-get update is unavoidable, default ubuntu docker image seems to have no state of apt repositories, even for basic stuff. I've pinned the image though.

.github/workflows/protobuf.yml
```diff
@@ -10,21 +10,12 @@ on:
...
-          ghc-version: '8.10'
-          cabal-version: 'latest'
-      - uses: ./.github/actions/asdf
-        with:
-          js: true
-          rust: true
-          golang: true
-      - run: ./dev/proto-generate.sh
-      - run: git diff --exit-code
+      - run: ./dev/proto-generate-in-docker.sh
```
keynmol, line 21: This is possibly a controversial change, but IMO the dockerised script uses the original one, so _both_ should work for CI to pass, we're not decreasing the coverage.

Development.md
```diff
@@ -27,10 +27,20 @@
...
    ```
    ./dev/proto-generate.sh
    ```
+
    For the Haskell bindings, see `bindings/haskell/README.md`.
+
+   You can also run the same script in a pre-configured Docker environment,
+   by running `./dev/proto-generate-in-docker.sh` - this command will
+   build the environment and run ./dev/proto-generate.sh script. The only
+   dependency you need is Docker.
```
varungandhi-src, line 42: Maybe we should recommend this first? Seems less error-prone?

.github/workflows/protobuf.yml
varungandhi-src: I think this is reasonable

Dockerfile.bindings
```diff
@@ -0,0 +1,45 @@
+FROM ubuntu@sha256:19478ce7fc2ffbce89df29fea5725a8d12e57de52eb9ea570890dc5852aac1ac
```
varungandhi-src, line 1: Maybe move this file to `dev/`? That's where all the other development scripts live. Unless that makes it harder to do some other things for some reason.

.dockerignore
```diff
@@ -0,0 +1 @@
+**/node_modules
```
varungandhi-src, line 1: Maybe copy stuff from `.gitignore` and mention that they should be kept in sync?

keynmol:
Things so far preventing this merge:

1. 
```
=== RUN   TestParse
&{0x100584358}
    binding_test.go:17:
                Error Trace:    /Users/antonsviridov/projects/sourcegraph/scip/cmd/scip/tests/reprolang/src/binding_test.go:17
                Error:          Received unexpected error:
                                cannot parse without language
                Test:           TestParse
--- FAIL: TestParse (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1005054b0]

goroutine 35 [running]:
testing.tRunner.func1.2({0x1005a5580, 0x1006d4ad0})
        /usr/local/go/src/testing/testing.go:1631 +0x1c4
testing.tRunner.func1()
        /usr/local/go/src/testing/testing.go:1634 +0x33c
panic({0x1005a5580?, 0x1006d4ad0?})
        /usr/local/go/src/runtime/panic.go:770 +0x124
github.com/sourcegraph/scip/cmd/scip/tests/reprolang/src.TestParse(0x140001329c0?)
        /Users/antonsviridov/projects/sourcegraph/scip/cmd/scip/tests/reprolang/src/binding_test.go:20 +0x120
testing.tRunner(0x140001329c0, 0x1005d8ee0)
        /usr/local/go/src/testing/testing.go:1689 +0xec
created by testing.(*T).Run in goroutine 1
        /usr/local/go/src/testing/testing.go:1742 +0x318
FAIL    github.com/sourcegraph/scip/cmd/scip/tests/reprolang/src        0.180s
FAIL
```

In reprolang tests – I tried syncing the manually embedded parser.h from treesitter so that versions match but it didn't work. I'll take a closer look, not familiar with this setup.

2. Some node-modules issues when running yarn in Docker.

.tool-versions
```diff
@@ -1,5 +1,6 @@
 golang 1.19.10
-nodejs 16.7.0
+nodejs 16.20.2
```
keynmol, line 2: Nodejs version in 16.x lineage that supports arm64 (for local development)

cmd/scip/tests/reprolang/package.json
```diff
@@ -12,6 +12,6 @@
     "nan": "^2.15.0"
   },
   "devDependencies": {
-    "tree-sitter-cli": "^0.20.4"
+    "tree-sitter-cli": "0.21.0"
```
keynmol, line 15: Minimal version that is both deterministic (https://github.com/tree-sitter/tree-sitter/issues/2755), available from arm64 (for local work), **and** available on NPM: https://www.npmjs.com/package/tree-sitter-cli?activeTab=versions (0.20.9 is missing)

keynmol: It's for tests only, so doesn't matter

dev/proto-generate.sh
```diff
@@ -19,7 +19,7 @@ fi
 
 echo "--- Haskell ---"
 command -v cabal > /dev/null 2>&1 || { echo >&2 "Haskell's 'cabal' command is not installed. Please install it first via https://www.haskell.org/ghcup/"; exit 1; }
-cabal install proto-lens-protoc-0.7.1.1 ghc-source-gen-0.4.3.0 --overwrite-policy=always --ghc-options='-j2 +RTS -A32m' --installdir="$PWD/.bin"
+cabal install proto-lens-protoc-0.8.0.1 ghc-source-gen-0.4.5.0 --overwrite-policy=always --ghc-options='-j2 +RTS -A32m' --installdir="$PWD/.bin"
```
keynmol, line 22: Bumps required because latest GHC doesn't support original versions...

package.json
```diff
@@ -6,9 +6,9 @@
   "author": "SCIP authors",
   "license": "MIT",
   "scripts": {
-    "build": "./node_modules/.bin/tsc --build --force bindings/typescript",
-    "prettier": "./node_modules/.bin/prettier --write --list-different '**/*.{ts,js(on)?,md,yml}'",
-    "prettier-check": "./node_modules/.bin/prettier --check '**/*.{ts,js(on)?,md,yml}'"
+    "build": "tsc --build --force bindings/typescript",
+    "prettier": "prettier --write --list-different '**/*.{ts,js(on)?,md,yml}'",
+    "prettier-check": "prettier --check '**/*.{ts,js(on)?,md,yml}'"
```
keynmol, line 11: No need to reference node_modules directly

.github/workflows/build-env-docker.yml
```diff
@@ -0,0 +1,69 @@
...
+        # It uses the `tags` and `labels` parameters to tag and label the image with the output from the "meta" step.
+      - name: Build and push Docker image
+        id: push
+        uses: docker/build-push-action@f2a1d5e99d037542a71f64918e516c093c6f3fc4
+        with:
+          file: dev/Dockerfile.bindings
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          platforms: linux/amd64,linux/arm64
```
keynmol, line 61: While this is slow (VERY slow), this workflow runs only on main and nothing depends on it - so it finishes when it finishes.

.github/workflows/protobuf-reprolang.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Log in to the Container registry
+        uses: docker/login-action@65b78e6e13532edd9afa3aa52ac7964289d1a9c1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - run: docker pull ghcr.io/sourcegraph/scip:latest || echo "no suitable cache"
```
kritzcreek, line 29: Why is it `scip-bindings-env` everywhere else, but just `scip` here?

kritzcreek (approved):
Using the container I can build all the bindings locally so I'm happy with this. Had one question about the image name

.github/workflows/protobuf-reprolang.yml
keynmol: Ah, last place where I forgot to update it.
Frankly I don't think this helps at all, but I will still update the image once I'm back at the computer

.github/workflows/protobuf-reprolang.yml
```diff
@@ -0,0 +1,37 @@
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - run: docker pull ghcr.io/sourcegraph/scip-bindings-env:latest || echo "no suitable cache"
+
+      - name: Regenerate protobuf bindings and reprolang parser
+        run: |
+          # We're changing the owner of the checkout folder to a particular user id,
+          # matching the user id of `asdf` user we create inside the docker container.
+          sudo chown -R 1001:1001 . && ./dev/generate-all-in-docker.sh
```
varungandhi-src, line 35: Is this needed because otherwise we might get permission errors inside the container? I don't understand why we need to create a separate `asdf` user inside the container.

.github/workflows/build-env-docker.yml
```diff
@@ -0,0 +1,69 @@
+# If this workflow stops working, first consult the documentation page where it was copied from.
+# https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages
+
+name: Create and publish a Docker image for bindings build environment
+
+# Configures this workflow to run every time a change is pushed to the branch called `release`.
+on:
+  push:
+    branches: ['main']
```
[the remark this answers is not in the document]
keynmol: I think with Github's own registry rate limiting is not really an issue, but I will at least limit the concurrency of this job to avoid producing intermediate docker images for commits that aren't top of branch.

.github/workflows/protobuf-reprolang.yml
keynmol: The reason for this is that I failed to make asdf work with a root user - there were some errors coming out from asdf itself which I didn't have time to debug.

It's an eyesore to have to do this, but to avoid it we'd need to make asdf work with root – which might not be worth the time, given the entire docker container is just a minimal hack to get back to reproducibility.

kritzcreek:
Hmm looks like the image publishing step failed on my PR after merging:

https://github.com/sourcegraph/scip/actions/runs/9691568032/job/26743288072

should we rip that out for now?

keynmol:
I will fix that workflow, I have a working one I can just copypaste.

PR #260, opened 2024-06-27
Use matrix build for multi-platform Docker build

keynmol:
This PR definitively fixes the multi-platform image generation, by using a matrix build.

### Test plan

- This workflow will run as part of main
- I also manually tested it – here's the published image(s): https://github.com/sourcegraph/scip/pkgs/container/scip-bindings-env


kritzcreek (approved):
Yolo

PR #277, opened 2024-09-21
Added Symbol Kinds for Dart

matthewnitschke-wk:
Adds support for 2 `SymbolKind` types that will be utilized by [scip-dart](https://github.com/Workiva/scip-dart/pull/146)

- `Extension` is for annotating dart's `extension` types, specifically the extension declaration, used for [resolving api conflicts](https://dart.dev/language/extension-methods#api-conflicts)

    ```dart
    extension Foo on String {
    //        ^^^ Kind.Extension
      void foo() => print('foo');
    }
    ```

    Open to suggestions on a different kind to use for this use case, or even a better name if `Extension` is to generalized 

- `Mixin` is for annotating dart [`mixin`s](https://dart.dev/language/mixins)
    
    ```dart
    mixin Foo {
    //    ^^^ Kind.Mixin
        int someField;
    }
    ```


varungandhi-src (approved):
It's OK to add language-specific cases to `Kind`. 👍🏽

PR #285, opened 2024-10-16
chore: Remove project board workflow

varungandhi-src:
We no longer use GitHub project boards so this isn't needed

### Test plan

n/a


kritzcreek (approved):

varungandhi-src:
PR to fix the prettier failure first here: https://github.com/sourcegraph/scip/pull/286

Issue #284, opened 2024-10-16
Please update protobuf dependency


RalfJung:
https://github.com/stepancheg/rust-protobuf/pull/737 fixes a bug in the protobuf crate where it wouldn't interact properly with build systems that use RUSTC_WRAPPER. This was released as a semver-compatible update, but this crate *pins* protobuf to `=3.2.0`, so sadly reverse dependencies can't pick up the new protobuf.

Would it be possible to update the protobuf dependency?

This is a blocker for https://github.com/rust-lang/rust/pull/127682.

varungandhi-src:
> Btw, after posting this issue I got a strange email about a failed pipeline

Created a PR to delete the job. https://github.com/sourcegraph/scip/pull/285

varungandhi-src:
The 3.7.1 version of the protobuf crate code doesn't compile with Rust 1.71.0 (version used in this repo), looking into upgrading that first.

```
error[E0445]: crate-private trait `MapFieldAccessor` in public interface
   --> /Users/varun/.local/share/mise/installs/rust/1.71.0/registry/src/index.crates.io-6f17d22bba15001f/protobuf-3.7.1/src/reflect/acc/v2/map.rs:115:1
    |
17  |   pub(crate) trait MapFieldAccessor: Send + Sync + 'static {
    |   -------------------------------------------------------- `MapFieldAccessor` declared as crate-private
...
115 | / pub fn make_map_simpler_accessor_new<M, T>(
116 | |     name: &'static str,
117 | |     get_field: for<'a> fn(&'a M) -> &'a T,
118 | |     mut_field: for<'a> fn(&'a mut M) -> &'a mut T,
...   |
121 | |     M: MessageFull,
122 | |     MapFieldAccessorImpl<M, T>: MapFieldAccessor,
    | |_________________________________________________^ can't leak crate-private trait

error[E0446]: private type `MapFieldAccessorImpl<M, T>` in public interface
   --> /Users/varun/.local/share/mise/installs/rust/1.71.0/registry/src/index.crates.io-6f17d22bba15001f/protobuf-3.7.1/src/reflect/acc/v2/map.rs:115:1
    |
33  |   struct MapFieldAccessorImpl<M, T>
    |   --------------------------------- `MapFieldAccessorImpl<M, T>` declared as private
...
115 | / pub fn make_map_simpler_accessor_new<M, T>(
116 | |     name: &'static str,
117 | |     get_field: for<'a> fn(&'a M) -> &'a T,
[...]

PR #287, opened 2024-10-17
chore: Bump rust-protobuf crates to 3.7.1

varungandhi-src:
We also bump the Rust version to 1.74.0 as older Rust
toolchains are unable to compile the protobuf dependency

Fixes https://github.com/sourcegraph/scip/issues/284

### Test plan

CI should pass


dev/proto-generate.sh
```diff
@@ -10,9 +10,9 @@ yarn --cwd ./bindings/typescript install --frozen-lockfile
 echo "--- cargo install rust-protobuf ---"
 PROTOC_GEN_RUST_VERSION="$(cat bindings/rust/Cargo.toml | grep 'protobuf =' | sed -E 's/.*\"=(.+)\"/\1/g')"
 if ! grep -q "$PROTOC_GEN_RUST_VERSION" "./.bin/PROTOC_GEN_RUST_VERSION" \
-   || ! test -f "./.bin/bin/protoc-gen-rust"; then
+   || ! test -f "./.bin/bin/protoc-gen-rs"; then
```
varungandhi-src, line 13: This update is because of the binary name change: https://github.com/stepancheg/rust-protobuf/blob/master/CHANGELOG.md

PR #291, opened 2024-11-29
chore: Use faster JSON encoding library

varungandhi-src:
`scip print --json` is a bit slow; this change provides a noticeable speedup.

E.g. on benchmarking against [flink-1.scip](https://drive.google.com/drive/folders/1z62Se7eHaa5T89a16-y7s0Z1qbRY4VCg).

```
❯ hyperfine '~/Code/scip/scip-before print --json flink-1.scip' '~/Code/scip/scip-after print --json flink-1.scip'
Benchmark 1: ~/Code/scip/scip-before print --json flink-1.scip
  Time (mean ± σ):      5.786 s ±  0.547 s    [User: 7.225 s, System: 0.861 s]
  Range (min … max):    5.499 s …  7.330 s    10 runs

Benchmark 2: ~/Code/scip/scip-after print --json flink-1.scip
  Time (mean ± σ):      3.323 s ±  0.197 s    [User: 4.294 s, System: 0.685 s]
  Range (min … max):    3.180 s …  3.870 s    10 runs

Summary
  '~/Code/scip/scip-after print --json flink-1.scip' ran
    1.74 ± 0.19 times faster than '~/Code/scip/scip-before print --json flink-1.scip'
```

### Test plan

n/a


cmd/scip/print_test.go
```diff
@@ -47,10 +47,10 @@ func TestJSONPrinting(t *testing.T) {
 
 	type JsonIndex struct {
 		Metadata struct {
-			ProjectRoot string `json:"projectRoot"`
+			ProjectRoot string `json:"project_root"`
```
varungandhi-src, line 50: This is technically a breaking change, but `scip print --json` is only meant to be used for debugging, so I think this is OK. This is coming about because in `scip.pb.go`, the struct tags specified for `protobuf`'s JSON encoding and the normal JSON encoding are different.

https://sourcegraph.com/github.com/sourcegraph/scip/-/blob/bindings/go/scip/scip.pb.go?L1488

```
	ProjectRoot string `protobuf:"bytes,3,opt,name=project_root,json=projectRoot,proto3" json:"project_root,omitempty"`
```

cmd/scip/print_test.go
```diff
@@ -63,7 +63,7 @@ func TestJSONPrinting(t *testing.T) {
 		log.Fatal(jsonErr)
 	}
 
-	require.Equal(t, roundtripResult.Metadata.ProjectRoot, "howdy")
-	require.Equal(t, len(roundtripResult.Documents), 1)
-	require.Equal(t, roundtripResult.Documents[0].RelativePath, "f")
+	require.Equal(t, "howdy", roundtripResult.Metadata.ProjectRoot)
+	require.Equal(t, 1, len(roundtripResult.Documents))
+	require.Equal(t, "f", roundtripResult.Documents[0].RelativePath)
```
varungandhi-src, line 68: I changed this order as expected values are supposed to be on the LHS.

PR #307, opened 2025-02-28
Refer to README with relative path instead of symlink in Rust crate

owickstrom:
### Test plan

N/A


kritzcreek (approved):
Verified via true and tested commit spamming

kritzcreek:
For context: this change is necessary, because the Github action that releases the Rust bindings to crates.io failed on main.

PR #311, opened 2025-03-06
CI: check if the cargo lock needs updating

owickstrom:
### Test plan

- Test in CI


kritzcreek:
Do we really want CI to fail on an unrelated PR, because some Rust dependency had an update in the meantime?

I'd think a lockfile really only makes sense when you're talking about a binary, whereas the bindings in here are used as a dependency, meaning the downstream application is going to pick the versions of transitive dependencies.

owickstrom:
@kritzcreek OK yeah, it might be too aggressive. The problem we had originally was the protobuf dependency constraint, and the version of this library itself, had changed and that wasn't reflected in the lockfile.

owickstrom:
@kritzcreek I'll close this for now. If the problem persists we can revisit a more balanced take.

kritzcreek:
SGTM, I think checking in the lock file for reproducible CI is probably a good idea. FWIW the lockfile in here does not affect what versions end up being picked in the downstream build in sg/sg.

PR #315, opened 2025-03-27
Allow IndexVisitor to return errors

efritz:
When processing an index in a streaming fashion, we're likely inserting into a database or doing something else that can fail and would likely benefit from a context value.

The way the Sourcegraph backend gets around this is to modify a scoped variable in the indexer. Implementations would be more clean if we just passed these signals around the visitor directly.


bindings/go/scip/parse_test.go
```diff
@@ -102,15 +106,22 @@ func checkRoundtrip(t *testing.T, index *Index) {
...
 
-	indexVisitor := IndexVisitor{func(metadata *Metadata) {
-		parsedIndex.Metadata = metadata
-	}, func(document *Document) {
-		parsedIndex.Documents = append(parsedIndex.Documents, document)
-	}, func(extSym *SymbolInformation) {
-		parsedIndex.ExternalSymbols = append(parsedIndex.ExternalSymbols, extSym)
-	}}
+	indexVisitor := IndexVisitor{
+		VisitMetadata: func(ctx context.Context, metadata *Metadata) error {
```
varungandhi-src, line 110: Let's use `_` if the parameter is not used.

In CHANGELOG.md, can you also add a Breaking change in the Unreleased section for Go bindings ?

bindings/go/scip/parse_test.go
```diff
@@ -3,13 +3,14 @@ package scip
...
 	"compress/gzip"
+	"context"
 	"io"
 	"os"
 	"regexp"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/gofuzz"
+	fuzz "github.com/google/gofuzz"
```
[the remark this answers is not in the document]
efritz: I don't think it's necessary, but it got added by Go tooling in VSCode.

bindings/go/scip/parse_test.go
efritz: Done!

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ Rust SCIP bindings:
   the latest release v3.7.2.
   (https://github.com/sourcegraph/scip/pull/313)
 
+Go bindings:
+
+- IndexVisitor now accepts a context object and returns an error.
```
varungandhi-src, line 14:
```suggestion
- BREAKING CHANGE: IndexVisitor now accepts a context object and returns an error.
  When an error is returned, the visitor short-circuits.
```

efritz: @varungandhi-src I think this was merged before the suggestion was committed.

efritz: Errr... actually I just misread.

Issue #233, opened 2024-03-06
SCIP sqlite database?


michaelpj:
I was taking a look at this project, it seems very cool. In particular, I think we could potentially use it as an intermediary format between the information files that the Haskell compiler GHC produces (HIE files) and the Haskell Language Server.

One thing we do at the moment with HIE files is that we index them and put them into a sqlite database (https://github.com/wz1000/HieDb). That way we can query things like references conveniently without having to maintain specialised data structures for our specific querying patterns.

Since SCIP is generic, I wonder if there could be a space for a generic SCIP sqlite schema and ingestor? Or is it just that everyone has sufficiently different application-specific needs that there is no point?

varungandhi-src:
I've had the same idea but primarily for the debugging use case (not as API) -- it'd be nice to be able to quickly convert a SCIP index to a SQL database to perform ad-hoc queries to check well-formedness, counts etc. Currently, you can do some scripting using the `scip print --json` functionality and then using JQ on top of that.

Right now, the `scip lint` subcommand does a bunch of custom tracking and traversals - it would be nice to specify those as SQL queries instead.

If we're making this a stable thing, it would require some care around maintaining backwards compatibility as we add new things (or explicitly stating that there aren't any compatibility guarantees, which would be frustrating through from a tooling POV).

sbalabanov:
There might be trouble creating a universal sqlite schema to suffice all of the language's requirements.
But there is also this: https://kythe.io/docs/kythe-storage.html
Which is another abstraction layer over sqlite.

PR #319, opened 2025-04-29
feat: Add conversion to SQLite (experimental)

varungandhi-src:
Adds a sub-command to convert a SCIP index to a SQLite database.

This is currently only meant for debugging; the schema should not be relied on.

Fixes https://github.com/sourcegraph/scip/issues/233

Fixes GRAPH-1200

### Test plan

Added simple smoke test


cmd/scip/convert.go
```diff
@@ -0,0 +1,601 @@
...
+
+const (
+	chunkSizeHint = 200 // Number of occurrences per chunk
+)
+
+func convertCommand() cli.Command {
+	var indexPath, outputPath, cpuProfilePath string
+
+	command := cli.Command{
+		Name:  "convert",
```
antonsviridov-src, line 32: `convert` seems like a very generic name, especially for an experimental command.

I suggest calling it `x-sqlite` – e.g. vcpkg adds `x-` prefix to the [commands and flags it deems experimental](https://learn.microsoft.com/en-us/vcpkg/commands/update-baseline) and not to be relied on

cmd/scip/convert_test.go
```diff
@@ -0,0 +1,160 @@
...
+	require.NoError(t, err)
+	defer func() { require.NoError(t, db.Close()) }()
+
+	writer, err := zstd.NewWriter(nil)
+	require.NoError(t, err)
+	converter := NewConverter(db, chunkSizeHint, writer)
+	err = converter.Convert(index)
+	require.NoError(t, err)
+
+	checkDocuments := func(db *sqlite.Conn) {
```
antonsviridov-src, line 55: I don't think we use this type of testing very often?
Consider extracting the setup code in a separate function and turning these cases into actual tests for IDEs to pick up

antonsviridov-src (approved):

cmd/scip/convert_test.go
varungandhi-src: > turning these cases into actual tests for IDEs to pick up

What do you mean by this? This command is not directly targeted for LSP/IDE use.

antonsviridov-src: Sorry, I meant for us, developers – `TestConvert_SmokeTest` gets picked up as an individual runnable test case, but those nested functions aren't.

It's not a big deal, just not a style I've seen used in our codebases a lot/

cmd/scip/convert.go
varungandhi-src: Changed the sub-command name to be `expt-convert`; users may not be familiar with the vcpkg convention on what `x-` means, whereas `expt` is more self-explanatory.

cmd/scip/convert_test.go
varungandhi-src: Broke up the logic into individual functions and sub-tests within the same test; GoLand will let you run sub-tests individually.

CleanShot 2025-05-09 at 15 27 33@2x

PR #327, opened 2025-08-19
feat: Upload kubernetes/kubernetes SCIP index

jupblb:
Adding only `kubernetes/kubernetes` for now, to see if this works well.

The Service Account used for SCIP index upload is named `scip-upload-service-account`. It has SCIP index upload permission and has repo permission for kubernetes/kubernetes.

### Test plan

N/A

### DIFF
--- a/.github/workflows/scip-examples.yaml
+++ b/.github/workflows/scip-examples.yaml
@@ -11,7 +11,6 @@ permissions:
 jobs:
   index-k8s:
     if: github.repository == 'sourcegraph/scip' # Skip running on forks
-    # runs-on: ubuntu-latest-4cores
     runs-on: ubuntu-latest
     timeout-minutes: 30
     concurrency:
@@ -26,8 +25,8 @@ jobs:
           ref: master
           fetch-depth: 1
 
-      # - name: Configure git safe.directory
-      #   run: git config --global --add safe.directory $GITHUB_WORKSPACE
+      - name: Configure git safe.directory
+        run: git config --global --add safe.directory $GITHUB_WORKSPACE
 
       - name: Get src-cli
         run: |


.github/workflows/scip-examples.yaml
```diff
@@ -0,0 +1,71 @@
...
+            -o /usr/local/bin/src
+          chmod +x /usr/local/bin/src
+
+      - name: Install Go
+        uses: actions/setup-go@v4
+        with: { go-version-file: 'go.mod' }
+
+      - name: Run scip-go
+        run: |
+          export GOWORK=off
```
varungandhi-src, line 52: Why are we setting this env var/what happens if we don't set it?

DESIGN.md
```diff
@@ -60,7 +60,6 @@ and then converting to SCIP for transmission.
     and zstd are already very good in terms of both compression
     speed and compression ratio.[^3]
 - Support efficient code navigation by itself.
-
```
varungandhi-src, line 63: This is an unrelated change.

varungandhi-src (changes requested):

.github/workflows/scip-examples.yaml
```diff
@@ -0,0 +1,71 @@
...
+          scip-go --version
+          scip-go --verbose
+
+      - name: Validate SCIP dump size
+        run: |
+          if [ $(stat -c%s index.scip) -lt 10000000 ]; then
+            echo "ERROR: SCIP dump suspiciously small (< 10MB)"
+            exit 1
+          fi
+          echo "SCIP dump size: $(du -h index.scip)"
```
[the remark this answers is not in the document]
jupblb: Will remember, changed.

.github/workflows/scip-examples.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Upload SCIP dump to Sourcegraph
+        run: |
+          src code-intel upload -no-progress \
+            -repo=github.com/kubernetes/kubernetes \
+            -file=index.scip
+        env:
+          SRC_ENDPOINT: https://sourcegraph.com/
+          SRC_ACCESS_TOKEN: ${{ secrets.SRC_ACCESS_TOKEN_DOTCOM_SCIP_SA }}
```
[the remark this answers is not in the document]
jupblb: I only added it to this repo as we will not upload SCIP indexes to dotcom from anywhere else. Should I move this to org-wide secrets?

.github/workflows/scip-examples.yaml
```diff
@@ -0,0 +1,71 @@
...
+    steps:
+      - name: Checkout kubernetes/kubernetes
+        uses: actions/checkout@v5
+        with:
+          repository: kubernetes/kubernetes
+          ref: master
+          fetch-depth: 1
+
+      - name: Configure git safe.directory
+        run: git config --global --add safe.directory $GITHUB_WORKSPACE
```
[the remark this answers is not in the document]
jupblb: I did try to run a workflow without it and it failed. I've seen this being used here: https://github.com/sourcegraph/sourcegraph/blob/f80ec0d84a54336daf66fa6e9c8a43cea9ac1fc2/.github/workflows/scip-go.yml#L26-L27

DESIGN.md
varungandhi-src: That's weird. Can you briefly look into why the check passed earlier? The CI has not been failing on `main`.

.github/workflows/scip-examples.yaml
varungandhi-src: Yeah, might as well index the full repo.

.github/workflows/scip-examples.yaml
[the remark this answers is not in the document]
varungandhi-src: Hmm, that's weird, I guess it's a relatively new git behavior.

https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9

Do you remember what error you saw? (I tried looking at previous Actions runs, but couldn't find it in a couple of minutes of looking.)

Fine with leaving this in, but it'd be nice to add a comment summarizing the error to describe why we have this line.

DESIGN.md
jupblb: Opened https://github.com/sourcegraph/scip/pull/328

PR #341, opened 2025-09-29
Fix go test and update dependencies

jupblb:
Running `go test ./...` on main ends up with the following error:
``` fish
~> go test ./...
# golang.org/x/tools/internal/tokeninternal
../../../.local/share/go/pkg/mod/golang.org/x/tools@v0.17.0/internal/tokeninternal/tokeninternal.go:78:9: invalid array length -delta * delta (constant -256 of type int64)
FAIL	github.com/sourcegraph/scip/bindings/go/scip [build failed]
ok  	github.com/sourcegraph/scip/bindings/go/scip/internal	0.990s
?   	github.com/sourcegraph/scip/bindings/go/scip/internal/shared	[no test files]
ok  	github.com/sourcegraph/scip/bindings/go/scip/memtest	1.261s
?   	github.com/sourcegraph/scip/bindings/go/scip/speedtest	[no test files]
ok  	github.com/sourcegraph/scip/bindings/go/scip/testutil	0.428s
FAIL	github.com/sourcegraph/scip/cmd/scip [build failed]
?   	github.com/sourcegraph/scip/cmd/scip/tests/reprolang/bindings/go/repro	[no test files]
ok  	github.com/sourcegraph/scip/cmd/scip/tests/reprolang/src	0.560s
FAIL
```

This was fixed mainly by updating `golang.org/x/tools`.

We explicitly require go v1.24.1 (instead of v1.24.0) because of [Sonic](https://github.com/bytedance/sonic/blob/48ba40b8e47bbb33828067f8250da9810f5b9b9f/README.md?plain=1#L10) requirement.

### Test plan
N/A


varungandhi-src (approved):
I think this is fine, but generally we support 2 major Go versions, so I think we shouldn't bump beyond this until Go 1.27 is released.

PR #356, opened 2026-02-23
Nixify project configuration

jupblb:
This is a seemingly large change that cleans up the build configuration for this repository and all its subprojects. In reality, most changes are trivial. Most importantly:

1. We stopped managing dependencies through various tools (docker, asdf, yarn, go.mod, ...) and just use `nix` for everything. This allows us to build packages and run checks using a single tool. flake.nix is the entry point.
2. Our CI now properly runs all the expected tests. Additionally, sample SCIP indexes are fetched from GCP. This helped uncover a bug in `bindings/go/scip/position.go`. Because we run more tests now, the CI times have not changed significantly but in the long run this opens up new opportunities to optimize our GitHub workflow setup.

To see how this works, first please [install nix](https://docs.determinate.systems/). Then you will be able to run the following commands:
- `nix develop` to enter shell that has all programs that are needed for development in its `PATH`.
- `nix build` to build `scip` binary.
- `nix flake check` to run checks.

### DIFF
--- a/checks.nix
+++ b/checks.nix
@@ -20,9 +20,6 @@
     preCheck = ''
       export SCIP_SAMPLE_INDEXES_DIR=${sampleIndexes}
     '';
-    # Limit parallel test execution to avoid OOM when loading large
-    # sample indexes concurrently in TestParseCompat.
-    checkFlags = [ "-parallel=2" ];
     installPhase = "touch $out";
   };
 
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -14,8 +14,18 @@ jobs:
 
   checks:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        check:
+          - go-bindings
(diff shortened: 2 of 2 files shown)


checks.nix
```diff
@@ -0,0 +1,103 @@
...
+      cd cmd/scip/tests/reprolang
+      cp -r src src-before
+      tree-sitter generate --abi 14
+      prettier --write 'src/grammar.json' 'src/node-types.json'
+      diff -rq src-before src
+    '';
+    installPhase = "touch $out";
+  };
+
+  rust-bindings = pkgs.rustPlatform.buildRustPackage {
```
chrapkowski-sg: How about `cmd/scip/tests/reprolang` ?

Previously we had:

```
      - run: cargo check
        working-directory: cmd/scip/tests/reprolang
```

chrapkowski-sg (approved):

PR #385, opened 2026-04-03
Introduce dedicated `SignatureDocumentation` message type

jupblb:
`SymbolInformation.signature_documentation` previously reused the `Document` message type, even though `Document` is designed to represent a source file on disk (with a "required" `relative_path`, etc.). This PR introduces a dedicated `SignatureDocumentation` message with a subset of `Document` fields.

The new message preserves the same protobuf field numbers as `Document` for the shared fields, so old and new indexes are binary wire-compatible in both directions. The `Document`-only field numbers are `reserved` in `SignatureDocumentation` to prevent accidental reuse.

This change is not compliant with Protobuf best practices (shouldn't change the type of an existing field) but we're also pre-1.0.

Supersedes https://github.com/scip-code/scip/pull/299


scip.proto
```diff
@@ -117,6 +115,25 @@ message Document {
   PositionEncoding position_encoding = 6;
 }
 
+// SignatureDocumentation represents the signature of a symbol as it's displayed
+// in API documentation or hover tooltips. It uses a subset of Document's fields
+// with the same field numbers for wire compatibility with older indexes that
+// encoded signatures using the Document message type.
+message SignatureDocumentation {
```
CatherineGasnier: Why not just Signature? I have a hunch the name `signature_documentation` was just a way to justify using the `Document` message, but really it's just a "signature"

scip.proto
CatherineGasnier, line 88: I wonder if some indexers were abusing this field to link a symbol to its containing file? In general, do we usually do bidirectional linking between messages? Right now a Document has a `symbols` field, but SymbolInformation does not point to its containing Document. Should it? Not super concerned by this but just raising.

scip.proto
```diff
@@ -117,6 +115,25 @@ message Document {
...
+  // The language of the signature, e.g. "java", "go", "python".
+  string language = 4;
```
CatherineGasnier: I'm wondering if this should be a field for Document only, to avoid the possibility of creating inconsistent data. But just nitpicking.

CatherineGasnier (commented):
Welcome change!

scip.proto
jupblb: There were two reasons I've left it here:
1. When we render hovers with signature in the Sourcegraph UI, we support syntax highlighting. It's easy to just get information about the language from this field. But it's not a problem at all to source it from the `Document` context.
2. Within boundary of a single language this is not a useful field, which is always the case right now. But if we were to support cross-language references _then_ it would make more sense, maybe.

I am fully ok with removing this field if you think there is no point in keeping it.

scip.proto
jupblb: I'm against having bidirectional linking and I think this convention fits the rest of the protocol definition. SCIP format is intended to be an intermediate form that is to be processed further. I know about only a single indexer that [sets](https://github.com/sourcegraph/scip-java/blob/4486a05471000ba4e784b72395ebf84207f24af8/scip-semanticdb/src/main/java/com/sourcegraph/scip_semanticdb/ScipSemanticdb.java#L284-L286) this field - but we mark it as `reserved` so it should not be a problem.

scip.proto
jupblb: Makes sense, renamed to `Signature`.

JamyDev (approved):

CatherineGasnier (approved):

jupblb:
Thank you! :)

PR #389, opened 2026-04-08
Add a check for GitHub Actions

jupblb:
Uses [action-validator](https://github.com/mpalmer/action-validator).

### DIFF
--- a/checks.nix
+++ b/checks.nix
@@ -6,10 +6,10 @@
   github-actions = pkgs.stdenv.mkDerivation {
     pname = "scip-github-actions";
     inherit version;
-    src = ./.github/workflows;
+    src = ./.;
     nativeBuildInputs = [ pkgs.action-validator ];
     buildPhase = ''
-      for f in *.yml *.yaml; do
+      for f in .github/workflows/*.yml .github/workflows/*.yaml; do
         [ -e "$f" ] && action-validator -v "$f"
       done
     '';
--- a/checks.nix
+++ b/checks.nix
@@ -3,6 +3,19 @@
   version,
 }:
 {
+  github-actions = pkgs.stdenv.mkDerivation {
+    pname = "scip-github-actions";
+    inherit version;
+    src = ./.github/workflows;
+    nativeBuildInputs = [ pkgs.action-validator ];
+    buildPhase = ''
+      for f in *.yml *.yaml; do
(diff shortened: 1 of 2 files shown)


checks.nix
```diff
@@ -3,6 +3,19 @@
   version,
 }:
 {
+  github-actions = pkgs.stdenv.mkDerivation {
+    pname = "scip-github-actions";
+    inherit version;
+    src = ./.github/workflows;
```
trly: `src = ./.github/workflows` is too narrow. action-validator uses `git ls-files` internally and will fail in the Nix sandbox because:
- No .git directory exists (git ls-files fails with git_ls_files_failed error)
- git is not in nativeBuildInputs
- Files referenced in path filters (e.g., scip.proto in proto-review.yaml) won't exist, causing glob_not_matched errors

Change this to just `src = ./.;`

## from workflows dir
```
cd .github/workflows && action-validator . -v
Unable to derive file name from src!
```

## from repo root
```
❯ action-validator .github/workflows/*.{yml,yaml} -v
Treating release.yml as a Workflow definition
Treating nix-checks.yaml as a Workflow definition
Treating proto-review.yaml as a Workflow definition
Treating scip-examples.yaml as a Workflow definition
```

checks.nix
```diff
@@ -3,6 +3,19 @@
...
+    nativeBuildInputs = [ pkgs.action-validator ];
+    buildPhase = ''
+      for f in *.yml *.yaml; do
```
trly: `for f in .github/workflows/*.yml .github/workflows/*.yaml;`

trly:
You'll also need to add `github-actions` to the matrix.check list in [nix-checks.yaml](https://github.com/scip-code/scip/blob/13d55c27687967b2fe763663e8f298f47383eebe/.github/workflows/nix-checks.yaml#L10-L11) of this will never run. Or is the intent to run this locally (I'm not super familiar with nix)

jupblb:
Thanks @trly, this is a very good observation!!

trly (approved):

checks.nix
jupblb: Already fixed

checks.nix
jupblb: Already fixed.

PR #390, opened 2026-04-10
Render external symbols snapshot outputs

jupblb:
Produce `external_symbols.txt` from `index.external_symbols`: https://github.com/scip-code/scip/blob/739faa9b609f0a86cea2be1b1d97d923b9baf9df/scip.proto#L28-L33

Example output (https://github.com/sourcegraph/scip-go/pull/198):
```
github.com/example/dep 0.1.test `github.com/example/dep`/T#
  relationship 0.1.test `sg/pr198`/Foo# implementation
```

### DIFF
--- a/bindings/go/scip/testutil/format.go
+++ b/bindings/go/scip/testutil/format.go
@@ -41,12 +41,9 @@ func FormatSnapshots(
 	for _, document := range index.Documents {
 		sourceFilePath := filepath.Join(localSourcesRoot, document.RelativePath)
 		snapshot, err := FormatSnapshot(document, commentSyntax, symbolFormatter, sourceFilePath)
-		err = symbolFormatter.OnError(err)
-		if err != nil {
+		if err = symbolFormatter.OnError(err); err != nil {
 			documentErrors = errors.Join(
-				documentErrors,
-				fmt.Errorf("%s: %w", document.RelativePath, err),
-			)
+				documentErrors, fmt.Errorf("%s: %w", document.RelativePath, err))
 		}
 		sourceFile := scip.NewSourceFile(sourceFilePath,
 			document.RelativePath,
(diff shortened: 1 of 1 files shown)


bindings/go/scip/testutil/format.go
```diff
@@ -192,6 +186,63 @@ func FormatSnapshot(
...
+	sort.Slice(sorted, func(i, j int) bool {
+		return sorted[i].Symbol < sorted[j].Symbol
+	})
+
+	formatSymbol := func(symbol string) string {
+		formatted, err := formatter.Format(symbol)
+		if err != nil {
+			return symbol
+		}
+		return formatted
```
trly, line 205: Is the intent to silently ignore malformed external symbols and malformed relationship target symbols with the external_symbols being added to external_symbols.txt however they're read?

bindings/go/scip/testutil/format.go
```diff
@@ -57,6 +57,16 @@ func FormatSnapshots(
 	if documentErrors != nil {
 		return nil, documentErrors
 	}
+
+	if len(index.ExternalSymbols) > 0 {
+		formatted := formatExternalSymbols(index.ExternalSymbols, symbolFormatter)
+		result = append(result, scip.NewSourceFile(
+			filepath.Join(localSourcesRoot, "external_symbols.txt"),
```
trly, line 61: Should this use RelativePath, similar to how it is used in [snapshot.go](https://github.com/scip-code/scip/blob/main/cmd/scip/snapshot.go#L83)?

bindings/go/scip/testutil/format.go
jupblb: Good point, we should use `symbolFormatter.OnError`!

bindings/go/scip/testutil/format.go
jupblb: `NewSourceFile` takes `absolutePath` as its first argument. This is fine.

bindings/go/scip/testutil/format.go
```diff
@@ -57,6 +57,19 @@ func FormatSnapshots(
 	if documentErrors != nil {
 		return nil, documentErrors
```
trly: Instead of returning early here, should we combine all formatting errors (external and document) and return them all at once?

If there are both document and external symbol errors, we won't know about them until the document errors are resolved.

jupblb: Good idea, changed :)

trly (approved):

PR #417, opened 2026-05-14
Combine renovate fix + nix-checks into one sequenced workflow

jupblb:
Renovate dep-bump PRs need post-processing (per-module `go mod tidy` and `nix-update` of vendor hashes) before the Nix checks can build them. Today this happens in two parallel workflows: the doomed nix-checks run starts at the same time as the fix and shows red until the fix re-triggers it.


trly (approved):
Branch protection rules will need to get updated to address the change from nix-checks-pass to ci-pass

PR #451, opened 2026-06-27
Add scip-cli to tools section

flesler:
Add [scip-cli](https://github.com/flesler/scip-cli) to the existing tools list in the README.

scip-cli is a Python CLI that builds on top of the `scip` binary from this repo (uses `scip expt-convert` for index conversion). It provides a token-efficient query interface and `analyze` command for AI agents — zero overlap with the core `scip` CLI, purely complementary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ For more details about indexers,
 see the [Sourcegraph documentation](https://docs.sourcegraph.com/code_navigation/references/indexers).
 
 Other tools which use SCIP include the [Sourcegraph CLI](https://github.com/sourcegraph/src-cli),
-and the SCIP CLI in this repo.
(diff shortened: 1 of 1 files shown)


jupblb:
The SCIP CLI implemented in this repository has an experimental [convert command](https://github.com/scip-code/scip/blob/main/cmd/scip/convert.go#L34) that outputs an SQLite database. I'd prefer if we work on that. Unless your project becomes relevant enough, I'm against listing it in this repo for now.

flesler:
Hi @jupblb, the project doesn't convert, maybe the wording was not clear. It uses this project's for conversion. What it does is it exposes a very clean interface and very optimized SQL queries to use data to collect on repos. Both for finding symbols, references, etc. While also exposing `analyze` to get a lot of insights into potential problems in a codebase

It has 0% overlap with this project, it's complimentary it builds on top

Issue #468, opened 2026-08-09
Would the scip-code org consider adopting the remaining indexers?


grinidx:
`scip-java`, `scip-go` and `scip-rust` moved to `scip-code` when SCIP's governance became independent. `scip-typescript`, `scip-python`, `scip-clang`, `scip-ruby` and `scip-dotnet` stayed under `sourcegraph`. I do not know whether that split was deliberate or simply where things stopped, so this is a question rather than a proposal.

The reason for asking is that the split has started to track activity. As of today:

| Repository | Org | Last push | Open issues and PRs |
|---|---|---|---|
| `scip-java` | scip-code | 9 Aug | |
| `scip-go` | scip-code | 9 Aug | |
| `scip` | scip-code | 6 Aug | |
| `scip-clang` | sourcegraph | 8 Aug | 46 |
| `scip-typescript` | sourcegraph | 7 Aug | 53 |
| `scip-dotnet` | sourcegraph | 1 Aug, no commit to `main` since 27 May | 33 |

That is not a complaint about anyone's priorities. Sourcegraph have been straightforward about them: on [scip-dotnet#61](https://github.com/sourcegraph/scip-dotnet/issues/61) a maintainer wrote that they have "no plans for any new features in SCIP indexers in general" and are "focusing on support, prioritising customer support". That is a reasonable position for a company to take, and it looks like exactly the situation the governance transition was designed for.

I have a concrete case, which prompted the question rather than being the point of it. `scip-dotnet` does not index Razor or Blazor. Those reach the compilation as source-generated documents, so iterating `project.Documents` misses every one of them: on a scaffolded app that is 0 documents for 7 `.cshtml` files, and 0 for 11 `.razor`. I have a patch, open as [scip-dotnet#117](https://github.com/sourcegraph/scip-dotnet/pull/117), with the existing snapshots byte-identical across net8.0, net9.0 and net10.0. It has not been reviewed, and four other pull requests there have been waiting longer, so I do not read that as anything personal.

I am not asking anyone to merge that patch. I am asking whether the adoption model already used three times might suit the indexers that were left, and if so what the process would be. If the answer is that the split was intentional and those repositories are staying put, that is a perfectly good answer and worth knowing.

I would be glad to help maintain the .NET indexer if that were useful. I have been working in that codebase, I have a fork with the Razor work in it, and I would rather contribute upstream than fork permanently.

[...]

PR #475, opened 2026-08-17
Add .NET bindings

grinidx:
Adds .NET bindings, as discussed in #468.

`bindings/` had go, haskell, java, kotlin, rust and typescript, so anyone writing a .NET SCIP producer or consumer vendors `scip.proto` by hand today. Sourcegraph's own `scip-dotnet` does, and so does the tool this came out of.

## The three things asked for

**The bindings.** Package id `Scip`, targeting `netstandard2.0`. Generated by a new `protoc_builtin: csharp` entry in `buf.gen.yaml`, alongside the existing `java` and `kotlin` ones. The generated file is committed and marked `linguist-generated`, matching what the other bindings do; CI regenerates and runs `git diff --exit-code`, so it cannot drift.

`netstandard2.0` because the bindings are pure generated data types with no framework-specific surface, so it costs nothing and reaches .NET Framework 4.6.1 and above, Mono and Unity, and every modern .NET. `Google.Protobuf` ships a `netstandard2.0` assembly.

**Nix.** A `dotnet-bindings` check in `checks.nix` in the same shape as `go-bindings`, plus what the devShell needs.

```
$ nix flake check
all checks passed!

$ nix build .#checks.x86_64-linux.dotnet-bindings
share/nuget/source/scip/0.9.0/scip.0.9.0.nupkg
```

The `formatting` check still passes with these files in place.

**Release automation.** A `publish-dotnet-bindings` job in `release.yaml` in the shape of the existing per-binding publish jobs, plus a `dotnet-bindings.yaml` PR check modelled on `jvm-bindings.yaml`, including its validation that the version matches `cmd/scip/version.txt`.

## One wrinkle worth knowing about

protoc's C# backend emits a `Descriptor` **class** that does not compile: CS0542, because every generated protobuf type already carries a static `Descriptor` property. So `proto-generate` renames the class to `SymbolDescriptor`.

The message itself is untouched, and `SymbolDescriptor.Descriptor.FullName` still returns `scip.Descriptor`, so nothing about the wire format or the schema changes. It is also what `sourcegraph/scip-dotnet` already ships, so anyone moving between the two sees the same name.

## Verified against a real index

Rather than only checking it compiles, a console app consuming the packed `Scip.0.9.0.nupkg` read a real `scip-typescript` 0.4.0 index of 60,867 bytes:

```
tool         scip-typescript 0.4.0
projectRoot  file:///.../ScentVerdict.Mobile
documents    4
occurrences  702
symbols      161
```

Re-encoding produces a byte-identical 60,867 bytes, and the reparsed index `.Equals(original)`.

## Needs something at your end

**A `NUGET_API_KEY` Actions secret**, scoped to "Push new packages and package versions". The `Scip` id is unregistered, so the first release creates it - worth claiming from an org account rather than a personal one if you have a preference.

Nothing else: no nix, runner or branch-protection changes.

One maintenance note for whoever does the next release: `bindings/dotnet/deps.json` must be regenerated whenever `Google.Protobuf` moves, in lock-step with flake.nix's protoc. The command is documented in `checks.nix`, and renovate is pinned off that dependency so it cannot bump it silently.


.github/workflows/dotnet-bindings.yaml
```diff
@@ -0,0 +1,44 @@
+name: .NET bindings
+
+# The Nix `dotnet-bindings` check already builds and packs this project with
+# the pinned SDK. This workflow covers the toolchain a contributor without
+# Nix uses, and annotates the exact file when the package version drifts
+# from cmd/scip/version.txt.
+
+on:
+  pull_request:
+    paths:
```
jupblb: I'm ok with this running on every PR. I prefer we didn't have to maintain this list of `paths`.

flake.nix
```diff
@@ -78,6 +78,23 @@
...
                 buf generate
                 goimports -w ./bindings/go/scip/scip.pb.go
+                # protoc's C# backend emits `class Descriptor` containing a
+                # static `Descriptor` property, which C# rejects (CS0542:
+                # member names cannot be the same as their enclosing type).
+                # Rename the generated class - not the Protobuf message, so
+                # the descriptor pool still says `scip.Descriptor` - to
+                # SymbolDescriptor, the name used both by scip.proto's own
+                # comments and by the existing scip-dotnet bindings.
+                sed -i -E \
```
jupblb, line 98: This is awful and if it can't be done better with nix it's fine to leave as is. But I'll appreciate a lot if this can be replaced with something more sound and future-proof.

jupblb (changes requested):
Overall this looks good to me but we need to switch to Trusted Publishing. I don't want to update the API key every month (current NuGet limitation).

Please make the appropriate changes and I'll proceed with the setup on NuGet. Alternatively, I can invite you to the https://www.nuget.org/profiles/scip-code org and you can do this yourself. Take your pick.

Thank you for taking care of this, much appreciated! :)

flake.nix
jupblb: I'm not sure this is better. 😓

jupblb (approved):
LGTM! What's your NuGet username?

